PR #47, opened 2013-11-04
Add database operations.

mehcode:
``` python
from sqlalchemy_utils import database_exists, create_database, drop_database

# Check if the database referenced in a sqlalchemy URL exists.
assert not database_exists('postgres://postgres@localhost/db_test_sqlalchemy_util')

# Create that database.
create_database('postgres://postgres@localhost/db_test_sqlalchemy_util')
assert database_exists('postgres://postgres@localhost/db_test_sqlalchemy_util')

# Now drop it.
drop_database('postgres://postgres@localhost/db_test_sqlalchemy_util')
assert not database_exists('postgres://postgres@localhost/db_test_sqlalchemy_util')
```

The functions have full support for postgres, mysql, and sqlite. I don't know much about oracle or other db engines but it _should_ work there as well.

### DIFF
--- a/sqlalchemy_utils/functions/database.py
+++ b/sqlalchemy_utils/functions/database.py
@@ -9,7 +9,7 @@ def database_exists(url):
     """Check if a database exists.
     """
 
-    url = copy(url)
+    url = copy(make_url(url))
     database = url.database
     url.database = None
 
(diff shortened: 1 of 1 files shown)


kurtmckee:
It looks like the branch associated with this still exists. I recommend deleting the branch to reduce the number of stale branches in the repo.

PR #90, opened 2014-09-16
Add PreciseTimestamp type

### LINKED ISSUE #86: PreciseTimestamp type
MySQL does not support the same kind of precision than PostgreSQL does in timestamps. We use a custom timestamp type to store timestamps as float in MySQL instead.

https://git.openstack.org/cgit/stackforge/gnocchi/tree/gnocchi/indexer/sqlalchemy.py#n47

I wonder if it would make sense to port this to sqlalchemy-utils. If so I'd be happy to submit a patch.

jd:
Closes #86


jd:
Now that MySQL supports this it's likely not needed.

Issue #20, opened 2013-07-07
MoneyComposite


kvesteri:
Simple composite type wrapper around integer amount column and CurrencyType currency column.

http://docs.sqlalchemy.org/en/latest/orm/mapper_config.html#composite-column-types

https://github.com/poswald/python-money

Edit:

Python needs better money handling, something along the lines of:

https://github.com/RubyMoney/money

kurtmckee:
Closing due to inactivity.

Issue #19, opened 2013-07-07
CurrencyType


kvesteri:
Preferably this could use:

https://github.com/poswald/python-money

Edit: actually this could use https://github.com/limist/py-moneyed

However atm both python-money and py-moneyed are not using Babel. Rather they are both reinventing the wheel for currency formatting etc. I added a ticket for this: https://github.com/limist/py-moneyed/issues/22

PR #108, opened 2014-12-18
Initial thoughts on a money composite

mehcode:
//ref #19 #20 

---

Been looking at it.

``` python
>>> tab.cash
USD 10

>>> tab.cash *= Decimal("2.1")
USD 21.0

>>> tab.cash.amount
Decimal('21.0')

>>> tab.cash.currency
'USD'
```

Currently it looks pretty bad from a usability standpoint but it "works".


kurtmckee:
It looks like the branch associated with this still exists. I recommend deleting the branch to reduce the number of stale branches in the repo.

PR #115, opened 2015-01-06
Update database.py

orf:
Postgres server_version can be more than just a number, this fails if it is for instance "9.4rc1". This patch should fix this


kvesteri:
Hi thanks a lot for this pull request. Before merging we need at least one test for this.

orf:
Hey, I have no idea how to test this. The implementation checks information_schema.columns, so I guess you would need to execute the test on Postgres 9.1 and 9.3?

Maybe drop_database shouldn't be terminating connections anyway?

kvesteri:
I think drop_database should terminate connections (maybe this should be optional though). You can easily create tests for it by mocking the query result (you can use http://has207.github.io/flexmock/ for this)

Issue #116, opened 2015-01-14
Support Enum for ChoiceType


jmagnusson:
I really like the ChoiceType field, it would be nice though to have the values enforced on the db level as well with `sqlalchemy.types.Enum` as the underlying type. What do you think?

kvesteri:
Yes this would require the choices to be passed to the underlying type (if its Enum). I think this would be a nice feature :+1:

bouzlibop:
:+1: would be great to have that

vToMy:
I'm currently using: http://techspot.zzzeek.org/2011/01/14/the-enum-recipe/
But I think being able to map python `enum` type to `sqlalchemy.types.Enum` would be best.

jmagnusson:
To be clear the following works:

```python

from sqlalchemy import Column, Enum
from .db import Base

class MyModel(Base):
    STATUS_CHOICES = (
        ('published', 'Published'),
        ('hidden', 'Hidden'),
    )
    STATUSES = tuple(c[0] for c in STATUS_CHOICES)

    status = Column(
        ChoiceType(impl=Enum(*STATUSES, name='mymodel_status'), choices=STATUS_CHOICES),
        default='published',
        nullable=False
    )
```

I'm not sure there could be any more clean way to do this. Can you?

Perhaps passing the kwargs of `ChoiceType` to the Enum. Like this:

```python
ChoiceType(impl_class=Enum, name='mymodel_status', choices=STATUS_CHOICES)

# Would be turned into this:
impl = impl_class(*(c[0] for c in choices), **{'name': name})
ChoiceType(impl=impl, choices=choices)
```

dbanty:
I would very much like this to happen.  Currently if I have an `enum.Enum` that I want a model to use, I have these choices:

1. Use the SQLAlchemy Enum type which will store the value as an Enum in the backing database (in my case MySQL).  However, it will not type coerce to the Enum from a value for me which is very handy.
2. Use ChoiceType, get my type coercion, but lose the backing tech
3. Use the method pointed out by @jmagnusson  above which gets me what I want but requires a bunch of extra code

I plan on going with option 3 for now so I can both have my cake and eat it, but it would be nice to make this the default behavior.

charlax:
Looks like this is now supported?

https://sqlalchemy-utils.readthedocs.io/en/latest/data_types.html#module-sqlalchemy_utils.types.choice

PR #118, opened 2015-01-15
Add support for built-in enum or backported enum34.

tonyseek:
Please review it. Thanks.


mehcode:
I like where you're going with this @tonyseek 

However, I think that we should have only one kind of restricted multiple choice type. See: https://github.com/kvesteri/sqlalchemy-utils/blob/master/sqlalchemy_utils/types/choice.py

Perhaps `ChoiceType` could be upgraded to both use enumerations as the python representation and the database representation. See: #116

tonyseek:
@mehcode I have tried to implement it.

tonyseek:
@mehcode How about now?

tonyseek:
@mehcode @kvesteri Is there something I can do? :-)

kvesteri:
Thanks for the PR! :train: 

I added some minor style fixes. I think the database representation of ChoiceType should be kept as String / Unicode for now. We can add support for Enum later on the db side.

tonyseek:
:smile:

PR #125, opened 2015-02-06
WIP Add support for JSONB

MSusik:
* Replaces default JSONType implementation with JSONB.

### DIFF
--- a/sqlalchemy_utils/types/json.py
+++ b/sqlalchemy_utils/types/json.py
@@ -1,5 +1,13 @@
+"""JSONType definition."""
+
 from __future__ import absolute_import
+
+import six
+
 import sqlalchemy as sa
+from sqlalchemy.dialects.postgresql.base import ischema_names
+
+from ..exceptions import ImproperlyConfigured
 
 json = None
 try:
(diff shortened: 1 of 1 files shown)


MSusik:
Is there a way to parametrize tests?

I'd like to have tests inside `JSONTestCase` run twice, for both values of `binary` in `JSONType` constructor.

lsbardel:
Hi guys, any reason this is not getting merged?

adamhadani:
:+1:  on merging, could really use this!

kvesteri:
I'm struggling to find a way how to use JSON and JSONB comparators with this type (http://docs.sqlalchemy.org/en/latest/dialects/postgresql.html?highlight=jsonb#sqlalchemy.dialects.postgresql.JSONB). If we can't find a way I'd be inclined to remove the PostgreSQL dialect implementation and just stick with Unicode for all dialects. Without the comparators we lose much of the power of JSON and JSONB.

Is there a reason why you want to use JSONType over native JSONB type? I'm currently just using JSONB on all my projects (on the other hand I'm using PostgreSQL on every project).

MSusik:
> Is there a reason why you want to use JSONType over native JSONB type? I'm currently just using JSONB on all my projects (on the other hand I'm using PostgreSQL on every project).

The case that made me write this simple PR was that the software I was developing was supposed to work with different versions of Postgre.

kvesteri:
SQLAlchemy will have native JSON type as of 1.1. I'll deprecate and remove JSONType from SA-Utils at some point.

PR #140, opened 2015-05-05
Added python_type implementation to arrow, email, ipaddress, password, a...

kelvinhammond:
Implemented python_type on a few types. This makes it work with serialization libraries such as [marshmallow-sqlalchemy](http://marshmallow-sqlalchemy.readthedocs.org).

[python_type in SQLAlchemy docs](http://docs.sqlalchemy.org/en/latest/core/custom_types.html#sqlalchemy.types.TypeDecorator.python_type)


kvesteri:
Thanks for the PR! :train:

PR #147, opened 2015-05-29
Allowing String based Columns for `ChoiceType` if python's `Enum` class is used

raabf:
Actually you can only use python's native `Enum` class with an Integer Column.
This MR allows also using other Column Types like `ENUM`, `TEXT`, `CHAR` (string based).
Especially the `ENUM` type is quite intuitive for mapping to the `Enum` class.

By default it uses the `name` of the `Enum` inherited class variable as mapping to the column.
But you can define the value for the column also by the `dbvalue` attribute, which you can set on the `Enum` inherited class variables.

Edit: I check for `isinstance(impl, types.Integer)`. Well `types.Integer` is a base class for all Integer types, so it should works well.

Something like the following should work:

``` python
        import enum

        class UserType(enum.Enum):
            admin = 1
            regular = 2

        UserType.admin.dbvalue = 'administrator'

        class User(Base):
            __tablename__ = 'user'
            name = sa.Column(sa.Unicode(255))
            type = sa.Column(ChoiceType(UserType, impl=sa.Enum('administrator', 'regular')))


        user = User(type=1)
        user.type  # <UserType.admin: 1>

```

Additionally it corrects an small copy and paste error in the already existing doc of the module fb57bb9402e9004396a03a64c0a6c80e975898cf

**TODO** (Edit)
- [ ] Docs for the new features (lets first see if the MR is accepted)
- [ ] Testing? It works for me with an `ENUM` and a varchar column
- [ ] Code review
- [ ] Add Test for Travis CI (need a string column in the test DB for that)

### DIFF
--- a/sqlalchemy_utils/types/choice.py
+++ b/sqlalchemy_utils/types/choice.py
@@ -199,7 +199,7 @@ class ChoiceTypeImpl(object):
 
 
 class EnumTypeIntImpl(object):
-    """The implementation for the ``Enum`` usage."""
+    """The implementation for the ``Enum`` usage with Integer based DB type."""
 
     def __init__(self, enum_class):
         if Enum is None:
--- a/sqlalchemy_utils/types/choice.py
+++ b/sqlalchemy_utils/types/choice.py
@@ -205,7 +205,7 @@ class ChoiceTypeImpl(object):
 
 
 class EnumTypeIntImpl(object):
-    """The implementation for the ``Enum`` usage with Integer based DB type."""
+    """The implementation for the ``Enum`` usage with Integer based Columns."""
 
     def __init__(self, enum_class):
         if Enum is None:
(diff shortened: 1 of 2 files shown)


sqlalchemy_utils/types/choice.py
```diff
@@ -146,9 +146,16 @@ def __init__(self, choices, impl=None):
         if (
             Enum is not None and
             isinstance(choices, type) and
+            issubclass(choices, Enum) and
+            isinstance(impl, types.Integer)
```
raabf: Well `types.Integer` is a base class for all Integer types, so it should works well.

kurtmckee:
It appears that this PR is stale. It also appears that sqlalchemy-utils has support for a `ChoiceType` that may match some of this functionality.

I recommend closing this PR.

PR #151, opened 2015-06-08
Use shields.io instead of pypip.in.

tonyseek:
It seems that the pypip.in has been unreachable.

@kvesteri Please reivew it. Thanks.


kvesteri:
Thanks for the PR! :bus:

PR #152, opened 2015-06-12
ChoiceType: non-None should not return None

dset0x:
EnumTypeImpl would use `None` instead of `0` due to bad checking.


dset0x:
I will likely have to revisit my views on returning `None`. Marking this as WIP until I have the time.

dset0x:
I've updated this PR; reintroduced support for `None` (otherwise we're breaking the SQLAlchemy model, how silly of me!). Let me know what you think. Thanks!

kvesteri:
Thanks for the PR! :train:

PR #154, opened 2015-06-27
Adds custom Numeric type PriceType

dappiu:
Useful to store Price objects from [prices](https://github.com/mirumee/prices) package


kvesteri:
I would like this to be called MoneyType instead of PriceType (makes more sense). Also it would be great to investigate and add support for at least some other Money backend (for example https://github.com/limist/py-moneyed).

dappiu:
I agree PriceType is not a particularly appropriate name. I'll look into py-moneyed (another not too exciting name, imho:) and try to understand what I can do here.

PR #155, opened 2015-07-10
Some drivers convert uuid values to uuid.UUID by themselves

dahlia:
Some drivers (e.g. [py-postgresql][1]) convert PostgreSQL `uuid` values to Python `uuid.UUID` objects by themselves.

[1]: http://python.projects.pgfoundry.org/


kvesteri:
Thanks for the PR! :train:

PR #160, opened 2015-08-26
Duplicated line in setup.py

stephane:
This test is already written line 14


kvesteri:
Thanks for the PR! :train:

PR #163, opened 2015-09-08
Fix locale territory parsing

dahlia:
`LocaleType` had lost locale’s territory, for example, `Locale('ja', territory='JP')` becomes `Locale('ja_JP')`.  This patch fixes the bug.  (See also the added test.)


kvesteri:
Thanks for the PR! :train:

PR #165, opened 2015-09-16
Allow reproducible build

thomasgoirand:
With this patch, building the Debian package is now predictible
and always the same. See Debian bug report here:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=799206


kvesteri:
Thanks for the PR! :train:

PR #181, opened 2016-01-18
Use pytest fixtures to reduce complexity and repetition

jmagnusson:
Also:
* Allow override of database name and user in tests (important for me as I would have to mess with my PSQL and MySQL database users otherwise)
* Use `dict.items` instead of `six.iteritems` as it sporadically caused `RuntimeError: dictionary changed size during iteration` in Python 2.6 tests.
* Fix typo DNS to DSN
* Adds Python 3.5 to `tox.ini`
* Added an `.editorconfig`
* Import `babel.dates` in `sqlalchemy_utils.i18n` as an exception would be raised when using the latest versions of babel.


jmagnusson:
Seems like the .isort.cfg is not being picked up. I've compared with your settings in `postgresql-audit`, but I can't find why it's not working here...

kvesteri:
Thanks for your hard work, this has been on my todo list for a very long time! Looks good. I have yet no idea why its not reading the ``.isort.cfg`` file. If we can figure that out, I'll merge this.

jmagnusson:
I had to configure tox to install all project dependencies for it to work. isort would recognize third party modules as first party otherwise, thus wanting to group them with `sqlalchemy_utils` imports.

Two more changes:

* Added a fix for Python 2.6 as well as the tests would sporadically fail otherwise, with a  `RuntimeError: dictionary changed size during iteration`.
* Added an `.editorconfig`

kvesteri:
Thanks for the PR! :train:

jmagnusson:
Thanks for merging, now I can do some real PRs :wink:. Any chance you could also merge #182?

PR #188, opened 2016-02-12
JSONType: Allow specifying `encode_cls` and `decode_cls`

singingwolfboy:
This commit allows a developer using `JSONType` to specify a JSONEncoder and JSONDecoder subclass, to allow encoding and decoding arbitrary values with JSON. The canonical example is datetime objects, but any object can be encoded and decoded in this way.


kvesteri:
Thanks for he PR. I would like to merge this but considering SQLAlchemy will have its own multi-platform JSON type (by version 1.1), I think it's better that we deprecate the JSONType from SA-Utils.

PR #192, opened 2016-02-23
Add MySQL fractional second support to Timestamp

lnielsen:
See also #86 and #90 

This fix ensures that Timestamp model also stores the microseconds part on MySQL, so that it works like on other dialects.


lnielsen:
Closing this again as it requires MySQL 5.6+ (e.g. Travis provides v5.5)

PR #193, opened 2016-02-25
whitespace causes setup.py parsing failure on Ubuntu 14.04

jaredeh:
When trying to install OpenStack via devstack I noticed a failure in the pip install of SQLAlchemy-Utils.  I traced it back to the setup.py containing whitespace in some of the fields of extras_require which seemed to throw the parsing on the default python install of Ubuntu 14.04, or at least the python on a system half through installing devstack from a default Ubuntu 14.04.

Error when running "python setup.py":

error in SQLAlchemy-Utils setup command: 'extras_require' must be a dictionary whose values are strings or lists of strings containing valid project/version requirement specifiers.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,12 +41,12 @@ extras_require = {
     'arrow': ['arrow>=0.3.4'],
     'intervals': ['intervals>=0.2.4'],
     'phone': ['phonenumbers>=5.9.2'],
-    'password': ['passlib >= 1.6, < 2.0'],
+    'password': ['passlib>=1.6,<2.0'],
     'color': ['colour>=0.0.4'],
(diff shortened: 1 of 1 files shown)


sigmavirus24:
I think this was a problem with setuptools 20.2.1 which has been pulled from PyPI. 20.2.2 should fix this issue.

sigmavirus24:
(In other words, this pull request is unnecessary)

tiborsimko:
Confirmed. Still, it would be nice to have a consistent whitespace style in `setup.py` dependency versioning (`foo>1.6` vs `foo > 1.6`), so that SQLAlchemy-Utils is resilient to potential future setuptools troubles of this kind... The whitespace unit testing had better belong there :)

PR #197, opened 2016-02-28
Set label for TranslationHybrid columns to its property name

jmagnusson:
Fixes column not being available by key in an sqlalchemy result object.

### DIFF
--- a/sqlalchemy_utils/i18n.py
+++ b/sqlalchemy_utils/i18n.py
@@ -95,16 +95,27 @@ class TranslationHybrid(object):
             getattr(obj, attr.key)[locale] = value
         return setter
 
-    def expr_factory(self, attr):
+    def expr_factory(self, attr, prop):
         def expr(cls):
             current_locale = cast_locale_expr(cls, self.current_locale)
             default_locale = cast_locale_expr(cls, self.default_locale)
-            return sa.func.coalesce(attr[current_locale], attr[default_locale])
+            ret = sa.func.coalesce(attr[current_locale], attr[default_locale])
+            try:
+                label = [k for k, v in vars(cls).items() if v is prop][0]
+            except IndexError:
+                # Fall back to a non-aliased expression if the hybrid property
+                # is not found on the entity. This happens on aliased classes
+                # for example.
+                pass
+            else:
+                ret = ret.label(label)
+            return ret
         return expr
 
     def __call__(self, attr):
-        return hybrid_property(
+        prop = hybrid_property(
             fget=self.getter_factory(attr),
-            fset=self.setter_factory(attr),
-            expr=self.expr_factory(attr)
+            fset=self.setter_factory(attr)
         )
+        prop.expr = self.expr_factory(attr, prop)
+        return prop
--- a/tests/test_translation_hybrid.py
+++ b/tests/test_translation_hybrid.py
@@ -117,3 +117,17 @@ class TestTranslationHybrid(object):
             locale = sa.Column(sa.String)
 
         Article.name
+
+    def test_column_labeling(self, session, City):
+        expected = ("coalesce(city.name_translations -> 'fi', "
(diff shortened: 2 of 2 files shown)


jmagnusson:
The CI errors seem to be related to your intervals project, @kvesteri.

jmagnusson:
I noticed that this isn't working too well. For example the following example:

```python
class Country(Base):
    __tablename__ = 'country'
    name_translations = sa.Column(HSTORE)
    name = translation_hybrid(name_translations)

class City(Base):
    __tablename__ = 'city'
    name_translations = sa.Column(HSTORE)
    name = translation_hybrid(name_translations)
    country_id = sa.Column(sa.Integer, sa.ForeignKey(Country.id))
    country = relationship(Country)

session.query(City.name, Country.name)
```
Throws:
```python
sqlalchemy.exc.InvalidRequestError: Ambiguous column name 'coalesce(city.name_translations -> 'fi', city.name_translations -> 'en')' in result set! try 'use_labels' option on select statement.
```
The labels are not override:able with `.use_labels()` so it simply does not make sense.

I've been trying to find a way to mimic how SQLAlchemy automatically handles regular declarative attributes by prepending the table name. Using the same models as above we can see how the columns are aliased as `city_name_translations` and `country_name_translations`:

```python
>>> country = Country(name='Finland')
>>> city = City(name='Helsinki', country=country)
>>> session.add(city)
>>> session.commit()
>>> q = session.query(City.name_translations, Country.name_translations)
>>> print(q)
SELECT city.name_translations AS city_name_translations, country.name_translations AS country_name_translations
FROM city, country
>>> print(q.first()._asdict())
{'name_translations': {'fi': 'Finland'}}
```
Unfortunately I wasn't able to get this behavior for the `hybrid_property`. Thankful for any tips from the SQLAlchemy gurus out there. (Looking at you @kvesteri 😉)

kvesteri:
Before digging into this further: why do you want to automatically set labels for translation hybrids?

jmagnusson:
My use case is that I want to be able to just call `._asdict()` on the result items when I query separate attributes. They don't show up for translation hybrid attributes as can be seen in this example:

```python
import sqlalchemy as sa
import sqlalchemy.orm  # noqa
from sqlalchemy.ext import declarative
from sqlalchemy.dialects.postgresql import HSTORE
from sqlalchemy_utils import TranslationHybrid
from sqlalchemy_utils.functions import database_exists, create_database


def get_locale():
    return 'fi'


translation_hybrid = TranslationHybrid(
    current_locale=get_locale,
    default_locale='en'
)

Base = declarative.declarative_base()
Session = sa.orm.sessionmaker()


class Article(Base):
    __tablename__ = 'article'

    def __str__(self):
        return '<Article: {} - {}>'.format(self.name, self.content)

    name = sa.Column(sa.Text, nullable=False)
    content_translations = sa.Column(HSTORE)
    content = translation_hybrid(content_translations)

article = Article(name='My product', content='Lorem ipsum...')

dsn = 'postgresql://localhost/translationhybridtest'
engine = sa.create_engine(dsn, echo=True)

if not database_exists(dsn):
    create_database(dsn)
    engine.execute('CREATE EXTENSION hstore')

Session.configure(bind=engine)
Base.metadata.create_all(bind=engine)
session = Session()
session.add(article)
session.flush()

article = session.query(Article.name, Article.content).one()
print('name' in article._asdict())  # True
print('content' in article._asdict())  # False

article = session.query(Article.name, Article.content.label('content')).one()
print('name' in article._asdict())  # True
print('content' in article._asdict())  # True

```

kvesteri:
Hmm I'm not sure I want to change this functionality to make it work for a use case that uses a private method. Also I wouldn't know how to make it work - I would have to dig deep into SA internals.

kvesteri:
I'm closing this for now as the provided use case in my opinion does not indicate strong enough purpose for adding this feature.

PR #198, opened 2016-02-28
Ensure that queried entities are taken from the alias

jmagnusson:
Previous behavior was to implicitly join in the parent of the alias to the query, causing two joins when there should only be one.

### DIFF
--- a/tests/test_translation_hybrid.py
+++ b/tests/test_translation_hybrid.py
@@ -2,6 +2,7 @@ import pytest
 import sqlalchemy as sa
 from flexmock import flexmock
 from sqlalchemy.dialects.postgresql import HSTORE
+from sqlalchemy.orm import aliased
 
 from sqlalchemy_utils import i18n, TranslationHybrid  # noqa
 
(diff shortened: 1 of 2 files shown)


tests/test_translation_hybrid.py
```diff
@@ -117,3 +118,9 @@ class Article(Base):
             locale = sa.Column(sa.String)
 
         Article.name
+
+    def test_aliased_entities(self, session, City):
+        # Ensure that queried entities are taken from the alias
+        CityAlias = aliased(City)
+        query_str = str(session.query(CityAlias.name))
```
kvesteri: No need for session testing here. Why not just test the compilation of ``CityAlias.name`` expression.

```python
str(aliased(City).name.compile())
```

jmagnusson: Actually I'm making sure that City isn't joined twice when querying. I'll clarify the test method name.

kvesteri:
Thanks for the PR! :train:

PR #199, opened 2016-03-07
fixed error formatting in currency.py

vToMy:


kvesteri:
Thanks for the PR! :train:

PR #201, opened 2016-03-11
using the new import standard

vToMy:
helps detect package when using PyInstaller.


kvesteri:
Thanks for the PR! :train:

PR #207, opened 2016-03-28
Handle default sqlite database

RobertDeRose:
SQLAlchemy's documentation states a SQLite database with a default to an in-memory database. The docs say you only need to provide a URL of "sqlite://" to get this functionality.

`create_database`, `database_exists`, and `drop_database` all did not handle this, in addition `drop_database` improperly reference the `url.database` after if is set to `None`. This was changed to use the `database` reference like the rest of the database types.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -68,6 +68,11 @@ def sqlite_memory_dsn():
 
 
 @pytest.fixture
+def sqlite_none_database_dsn():
+    return 'sqlite://'
+
+
+@pytest.fixture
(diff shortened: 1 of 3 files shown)


RobertDeRose:
Addresses Issue https://github.com/kvesteri/sqlalchemy-utils/issues/208

kvesteri:
Thanks for the PR! :train:

RobertDeRose:
Cool thanks!

PR #210, opened 2016-04-15
Use the `postgresql_db_user` fixture in test_template tests

jmagnusson:
Before it was hard-coded to the `postgres` user.
I’m not quite sure why these tests failed for me, considering that the calls to the engine should be mocked. Perhaps a new version of `flexmock` has changed something. Downgrading SQLAlchemy did not help.


kvesteri:
Thanks for the PR! :train:

PR #211, opened 2016-04-20
Fix the LazyCryptContext stays lazy after init

olegpidsadnyi:
The problem is that the constructor of the PasswordType immediately accesses the context inside the constructor which makes it not lazy anymore.
I added a test to prevent that in the future.


kvesteri:
Thanks for the PR! :train:

Issue #216, opened 2016-04-29
Call for opinion: shouldn't IPAddressType be validated on the way in?


lingfish:
Should the `six.text_type(value)` call (https://github.com/kvesteri/sqlalchemy-utils/blob/48245b14c773009cbfea571cf3f9edef05f25224/sqlalchemy_utils/types/ip_address.py#L62) perhaps have a call to `ip_address()` inside it, so that it throws an exception if it's invalid?

Something like:

```python
return six.text_type(ip_address(value)) if value else None
```

I'm just thinking that the expectation with most other SA types is that "something" (usually the DB itself), would reject invalid data (rather than the app using it), and this type certainly does on the way out (https://github.com/kvesteri/sqlalchemy-utils/blob/48245b14c773009cbfea571cf3f9edef05f25224/sqlalchemy_utils/types/ip_address.py#L65) implied by the `ip_address()` call.

I discovered this because I could erroneously insert "1.1.1.1 255.255" into a column, but then when reading it back out via `eve-sqlalchemy`, I get a ValueError:

```python
  File "/srv/virtualenv/something/lib/python2.7/site-packages/sqlalchemy_utils/types/ip_address.py", line 65, in process_result_value
    return ip_address(value) if value else None
  File "/srv/virtualenv/something/lib/python2.7/site-packages/ipaddress.py", line 166, in ip_address
    address)
ValueError: u'10.64.21.33 255.255.' does not appear to be an IPv4 or IPv6 address
```

kvesteri:
If you are using [automatic data coercion](http://sqlalchemy-utils.readthedocs.io/en/latest/listeners.html#automatic-data-coercion) then the ip address already gets validated when setting the value to the column, which is even better as it gets validated earlier. Thus I don't think its needed to validate this in the parameter binding phase.

lingfish:
Nice, thanks. I'm going to throw you a PR with a slight update the doco then.

PR #217, opened 2016-05-04
Expand out the doco around auto coercion and validation

lingfish:
As per https://github.com/kvesteri/sqlalchemy-utils/issues/216


kvesteri:
Thanks for the PR! I did a small change and merged this manually.

PR #219, opened 2016-05-16
Wrap PhoneNumber parse exception so SQLAlchemy won't wrap/swallow it

JackWink:
Currently, if any type throws an exception that doesn't inherit from SQLAlchemy's [`DontWrapMixin`](http://docs.sqlalchemy.org/en/latest/core/exceptions.html#sqlalchemy.exc.DontWrapMixin) then SQLAlchemy wraps the error as a `StatementError`.

This PR wraps the `NumberParseException` from `python-phonenumbers` in a custom exception using the `DontWrapMixin` so we retain the original error message as well as a unique type for this exception. If this approach seems acceptable, I can work on wrapping the other types as well.

### DIFF
--- a/sqlalchemy_utils/__init__.py
+++ b/sqlalchemy_utils/__init__.py
@@ -82,6 +82,7 @@ from .types import (  # noqa
     Password,
     PasswordType,
     PhoneNumber,
+    PhoneNumberParseException,
     PhoneNumberType,
     register_composites,
     remove_composite_listeners,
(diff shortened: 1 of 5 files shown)


kvesteri:
Before merging this needs at least one unit test.

JackWink:
On second thought, it might make more sense to have `PhoneNumberParseException` extend `NumberParseException` if that would mean the handling code can still try to catch the `NumberParseException` or this named exception, thoughts?

I'll  play around with it tomorrow and write a test case, thanks for the quick response!

kvesteri:
Yes I agree that would make more sense.

JackWink:
👍 should be good for a second check

kvesteri:
Thanks for the PR! I made some small changes and merged this manually.

PR #220, opened 2016-05-20
Fix for phonenumber test failing (unicode issue)

JackWink:
A different exception was being thrown in py27 since we passed in
non-unicode characters for the name field.


kvesteri:
Thanks for the PR! :train:

PR #221, opened 2016-05-22
Fix issue with PasswordType & Flask-Migrate

osya:
When executing command `python manage.py in model contains PasswordType field the following error appeares: "raise KeyError("unknown CryptContext keyword: %r" % (key,))
KeyError: "unknown CryptContext keyword: 'length'""

### DIFF
--- a/sqlalchemy_utils/types/password.py
+++ b/sqlalchemy_utils/types/password.py
@@ -157,6 +157,9 @@ class PasswordType(types.TypeDecorator, ScalarCoercible):
 
         self._max_length = max_length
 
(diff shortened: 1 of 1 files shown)


kvesteri:
Thanks for the PR! Could you add a simple test for this as well?

osya:
I find an possible issue (kwargs not passed). And I will do the same __repr__ for TimezoneType. So I close this PR and create new PR for both changes. According to test - I am nnew to tox, tox tests in sqlalchemy-utils are not passed for me for some reason. So I will investigate and write tests but in future.

PR #225, opened 2016-06-10
Document that force_auto_coercion() is required for PasswordType

singingwolfboy:
I wasted a *lot* of time trying to track down why this wasn't working in my project. Hopefully, I can prevent other people from wasting their time, as well.


kvesteri:
Thanks for the PR! :train:

PR #227, opened 2016-06-23
doc update: cPython2.6 dropped since 0.32

guillp:
Updated Platform support doc


kvesteri:
Thanks for the PR! :train:

Issue #229, opened 2016-07-04
EmailType ignores length argument


tvuotila:
EmailType ignores length argument. For example compare the two:
```
>>> import sqlalchemy
>>> text_column = sqlalchemy.Unicode(length=75)
>>> text_column.length
75
>>> import sqlalchemy_utils
>>> email_type = sqlalchemy_utils.EmailType(length=75)
>>> email_type.length 
255
```

This breaks everything that depends on type length like [automatic length validator](https://github.com/kvesteri/wtforms-alchemy/blob/c1a22855fd2a99ba02458c93a936359ef6b60a80/wtforms_alchemy/generator.py#L580) or alembic revision autogeneration (`alembic revision --autogenerate`).

PR #230, opened 2016-07-05
Allow setting custom length to EmailType, refs #229

quantus:
Ticket #229.

This change allows users to specify the exact length the email field should have in database. By default the length is as it was before.

### DIFF
--- a/sqlalchemy_utils/types/email.py
+++ b/sqlalchemy_utils/types/email.py
@@ -4,8 +4,11 @@ from ..operators import CaseInsensitiveComparator
 
 
 class EmailType(sa.types.TypeDecorator):
-    impl = sa.Unicode(255)
+    impl = sa.Unicode
     comparator_factory = CaseInsensitiveComparator
+
+    def __init__(self, length=255, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


sqlalchemy_utils/types/email.py
```diff
@@ -4,9 +4,12 @@
 
 
 class EmailType(sa.types.TypeDecorator):
-    impl = sa.Unicode(255)
+    impl = sa.Unicode
     comparator_factory = CaseInsensitiveComparator
 
+    def __init__(self, length=255):
```
tvuotila: I would add `*arg, **kwargs` pass them to the next `__init__`. Missing `**kwargs` prevents arguments `collation`, `convert_unicode`,` unicode_error` and `_warn_on_bytestring`. Missing `*args` hinders subclassing.

quantus: I was also thinking about that, but then I just couldn't find any compelling reason to have those options there. On the other hand I don't see any negative sides in passing those extra variables to the parent class, so I may as well do that.

I'll update this PR tomorrow.

PR #231, opened 2016-07-14
Feature/multi column observer

quantus:
This PR adds support for multi column observers.

Things to consider:
 - Manually specifying `observer` to `observes` happens in a different way after this change. This can't be avoided with the chosen API as we support Python 2, that doesn't support [PEP 3102](https://www.python.org/dev/peps/pep-3102/).
 - Is there need for further documentation?
 - Is there need for further testing? If so, what should be tested?


kvesteri:
Thanks for the PR! :train:

Issue #40, opened 2013-08-23
Make batch_fetch support custom query constructs for fetched attribute paths


kvesteri:
This would enable for example ordering of batch fetched relationships.

Issue #106, opened 2014-12-15
Support alembic for the library of custom types


mehcode:
> https://groups.google.com/d/msg/sqlalchemy-alembic/oBAY2K1Xkeo/a0kMlVPdfuwJ
> ---
> 
> Alembic gives you _two_ options to determine how a type object is rendered when using autogenerate.  The first is to implement **repr**() on your type, however that won’t give you access to the parent module.   The second is to implement render_item(), which gives you complete control over how any specific type is rendered.  Implementing this will give you a way to fully control for any specific type how it is rendered.  this is configurable via render_item: http://alembic.readthedocs.org/en/latest/api.html#alembic.environment.EnvironmentContext.configure.params.render_item

---

It looks like we either need to proxy `repr` to the underlying `impl` or implement `render_item` (the former seems easier).

bx2:
Both solutions would not work, because of the [recent bug in Alembic](https://github.com/zzzeek/alembic/pull/17), which is now fixed.

Of course there is still the problem of representation. If you auto-generate from Alembic, then your fields are rendered with types from sqlalchemy-utils, but with errors i.e. UUIDType has no values passed beside the length, which causes an error. Same goes for ChoiceType etc.

mehcode:
Nah.. the idea is to generate with the underlying `impl`. Type decorators aren't meant to go in alembic (from what I understand). 

So a `UUIDType` would ouput a `BINARY` or `CHAR` depending on the dialect and options.

I've done a quick sketch of this and it works, I just haven't had time to do it in full yet.

frol:
Isn't this issue annoying absolutely everybody?!

dbrumley:
Yes, it is annoying me.  One specific case is for `PasswordType`.  Any objection to a PR that fixed this one; it's simple with:
```
   def __repr__(self):
        return "PasswordType(max_length=%r)"%self._max_length
```

hsoto92:
Still having trouble with `UUIDType`. 

```
File "C:\www\liu-passport-api\flask_application\migrations\versions\88f9317a1de1_.py", line 151, in upgrade
    sa.Column('user_id', UUIDType(length=16), nullable=True),
TypeError: <flask_script.commands.Command object at 0x041F26D0>: __init__() got an unexpected keyword argument 'length'
```

Any idea how to fix this?

eshlox:
[...]

PR #233, opened 2016-07-21
Fix for bug #40 for Password type (only)

dbrumley:
Fix makes representation consistent with constructor call by printing
length as the expected max_length parameter. Note this helps downstream
alembic migrations work properly too.

### DIFF
--- a/sqlalchemy_utils/types/password.py
+++ b/sqlalchemy_utils/types/password.py
@@ -169,6 +169,10 @@ class PasswordType(types.TypeDecorator, ScalarCoercible):
 
         return self._max_length
 
+    def __repr__(self):
+        # Ensure representation uses max_length, not length
+        return "PasswordType(max_length=%r)" % self.length
+
     def calculate_max_length(self):
         # Calculate the largest possible encoded password.
         # name + rounds + salt + hash + ($ * 4) of largest hash


kvesteri:
Thanks for the PR. Before merging this needs unit test(s).

frol:
This doesn't work for me when I use SQLite as a backend as it somehow uses `sa.NUMERIC(precision=128)`, and I still receive a bloated autogenerated migration:

```python
    op.alter_column('user', 'password',
               existing_type=sa.NUMERIC(precision=128),
               type_=sqlalchemy_utils.types.password.PasswordType(max_length=128),
               existing_nullable=False)
 ```

**Comment by dbrumley** [comment]
What doesn't work? The patch?  I didn't put a unit test, so there was never a PR that incorporated the patch.

**Comment by frol** [comment]
I applied the patch locally, and it didn't help me.

**Comment by frol** [comment]
Oh, yeah, this patch at least helps to produce a correct `type_` value using `max_length=` argument instead of the autogenerated `length=`, but Alembic's autogeneration still detects a change where it shouldn't:

```
Detected type change from NUMERIC(precision=128) to PasswordType(max_length=128) on 'user.password'
```

dbrumley:
ah, that i did not look into.

frol:
It seems that I wrongly assumed that #106 refers to that SQLAlchemy-Utils fields are getting into every new Alembic migration, at least that is what I experience using SQLite for prototyping. It turned out that the only raised concern was that incorrect `length=` argument passed instead of `max_length=`. I am sorry for hijacking your PR. While it addresses one problem, it doesn't (and, probably, shouldn't) address the SQLite-specific one, so I dug into this issue myself and as a result, I think the bug is in SQLAlchemy, so I reported it there: https://groups.google.com/forum/#!topic/sqlalchemy/GpkSJ0XUgwo

P.S. @dbrumley Why does your commit point to the issue #40? I think, it should point to #106.

PR #234, opened 2016-07-23
Added support for backref tuple to auto_delete_orphans

vToMy:
When using `sqlalchemy.orm.backref` it returns a tuple.


kvesteri:
Thanks for the PR, before merging this needs at least one unit test.

vToMy:
Done!
I just copied the test for the regular use (made sure it failed first, and that my change made it pass).
It would have probably been better to somehow run the same test for both classes, but I'm not sure how you can do this.

vToMy:
Hey,
Improved the test so there's no duplication.
Hopefully there's nothing to stop you from merging now :)

kvesteri:
Thanks for the PR!

PR #235, opened 2016-07-25
json: use dialect (de)serializer to encode JSON data

jd:


kvesteri:
Thanks for the PR. As with other PRs before merging this needs unit tests.

jd:
@kvesteri completely agree. In this case, this would require to change `connect()` in fixture to pass a custom JSON serializer AFAIU. Would it be acceptable?

kvesteri:
No need for that. You can just override the connection fixture on that test file / class.

kvesteri:
JSONType will be deprecated now that SQLAlchemy has its own generic JSON type. Thus I'm closing this PR and all other JSONType related PRs

PR #242, opened 2016-09-07
Add support for database dialects

hackermd:
Make checks for dialect names less restrictive to support potential custom dialects as described
in the [SQLAlchemy docs](http://docs.sqlalchemy.org/en/latest/core/connections.html#registering-new-dialects).


kurtmckee:
This PR looks like it has value, but it has gone stale in the last five years and doesn't have unit tests.

I recommend closing this PR.

PR #243, opened 2016-09-09
postgres connection url not being parsed correctly 

tpietsch:
postgres url parsing is removing the database from the url which cause problems when using the server url + "/databasename" method of connecting

it is also only an issue when using the "postgres" head of the connection string because this check only looks for the version "postgresql"

I have updated the url.database to be the database that is parsed from the URL that way the check exist function isn't looking at the "template1" database which the server running this script may not have access to read.

after reading postgres docs connecting to template1 makes sense. Looking into other ways since my current  situation disallows me to connect to template1.

### DIFF
--- a/sqlalchemy_utils/functions/database.py
+++ b/sqlalchemy_utils/functions/database.py
@@ -453,8 +453,8 @@ def database_exists(url):
 
     url = copy(make_url(url))
     database = url.database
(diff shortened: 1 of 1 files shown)


tpietsch:
@kvesteri looks like the build is failing cause the test itself was never connecting and now its actually trying to connect and cannot. Can you verify that when you have a sec.

Issue #245, opened 2016-09-27
PasswordType makes object unhashable in python 3


danqing:
I may be missing something obvious here, but I'm having the following issue:

```py
>>> u = User()
>>> u.password = 'password'
TypeError: unhashable type: 'User'
```

where `password` is a `PasswordType` field:

```py
password = Column(PasswordType(schemes=['pbkdf2_sha512']))
```

This is the full trace:

```py
/env/lib/python3.5/site-packages/sqlalchemy/orm/attributes.py in __set__(self, instance, value)
    222     def __set__(self, instance, value):
    223         self.impl.set(instance_state(instance),
--> 224                       instance_dict(instance), value, None)
    225
    226     def __delete__(self, instance):

/env/lib/python3.5/site-packages/sqlalchemy/orm/attributes.py in set(self, state, dict_, value, initiator, passive, check_old, pop)
    699         if self.dispatch.set:
    700             value = self.fire_replace_event(state, dict_,
--> 701                                             value, old, initiator)
    702         state._modified_event(dict_, self, old)
    703         dict_[self.key] = value

/env/lib/python3.5/site-packages/sqlalchemy/orm/attributes.py in fire_replace_event(self, state, dict_, value, previous, initiator)
    708                 state, value, previous,
    709                 initiator or self._replace_token or
--> 710                 self._init_append_or_replace_token())
    711         return value
    712

/env/lib/python3.5/site-packages/sqlalchemy/ext/mutable.py in set(target, value, oldvalue, initiator)
    472                 value = cls.coerce(key, value)
    473             if value is not None:
--> 474                 value._parents[target.obj()] = key
    475             if isinstance(oldvalue, cls):
    476                 oldvalue._parents.pop(target.obj(), None)

/env/lib/python3.5/weakref.py in __setitem__(self, key, value)
    376
    377     def __setitem__(self, key, value):
--> 378         self.data[ref(key, self._remove)] = value
    379
    380     def copy(self):

TypeError: unhashable type: 'User'
```

danqing:
It looks like the reason is because [if `__eq__` is defined but not `__hash__` the object becomes unhashable](http://stackoverflow.com/questions/1608842/types-that-define-eq-are-unhashable-in-python-3-x).

https://github.com/kvesteri/sqlalchemy-utils/blob/b1f84e5998730d356c57bfb145c9b95e9a9f6c04/sqlalchemy_utils/types/password.py#L45

danqing:
[...]

PR #246, opened 2016-09-27
Add hash function for PasswordType

danqing:
Trying to get a stab at #245


danqing:
Hmmm.... I'm not sure what's the right way to do it:

1. Objects that compare equal should have the same hash.
2. If that means encrypted "password" should hash to the same value as the string "password" then there's no point of encrypting.
3. And I don't know how to achieve 2 anyway. Unless we store the string somewhere..

PR #248, opened 2016-10-09
PhoneNumberType().python_type should be a class

fuhrysteve:
See:
http://docs.sqlalchemy.org/en/latest/core/type_api.html#sqlalchemy.types.TypeEngine.python_type

I found this as a result of a change introduced in marshmallow-sqlalchemy https://github.com/kelvinhammond/marshmallow-sqlalchemy/commit/12da838c8ca3e02516dd83852cef870d1d596b1b


kvesteri:
Thanks for the PR, I made some small alterations and added this.

PR #250, opened 2016-10-17
Introduce generic_repr() model decorator, that adds generic __repr__() to an ORM model.

AndrewPashkin:
Subj.

### DIFF
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -8,3 +8,11 @@ Timestamp
 .. module:: sqlalchemy_utils.models
 
 .. autoclass:: Timestamp
+
+
+generic_repr
+------------
+
+.. module:: sqlalchemy_utils.models
+
+.. autofunction:: generic_repr
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -1,21 +1,20 @@
+import sys
 from datetime import datetime
 
 import pytest
 import sqlalchemy as sa
 
-from sqlalchemy_utils import Timestamp
-
-
-@pytest.fixture
-def Article(Base):
-    class Article(Base, Timestamp):
-        __tablename__ = 'article'
-        id = sa.Column(sa.Integer, primary_key=True)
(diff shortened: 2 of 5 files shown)


AndrewPashkin:
@kvesteri, tests are green.

kvesteri:
Thanks for the PR! I think more pythonic way would be to add this as a class decorator (in the same fashion as https://docs.python.org/3.6/library/functools.html#functools.total_ordering).

AndrewPashkin:
@kvesteri, it makes sense.

AndrewPashkin:
@kvesteri, I've updated the PR.

docs/models.rst
```diff
@@ -8,3 +8,11 @@ Timestamp
 .. module:: sqlalchemy_utils.models
 
 .. autoclass:: Timestamp
+
+
+GenericReprMixin
```
kvesteri: The docs should refer to ``generic_repr``

tests/test_models.py
```diff
@@ -38,3 +37,52 @@ def test_updated(self, session, Article):
         session.commit()
 
         assert article.updated >= then and article.updated <= datetime.utcnow()
+
+
+class TestGenericReprMixin:
```
kvesteri: Rename this test case (its testing generic_repr not GenericReprMixin so maybe TestGenericRepr would be good).

frol:
I really like this idea! I hope this will be merged soon.

@AndrewPashkin Could you, please, address the @kvesteri's requests? Those seem to be so small renamings after all.

AndrewPashkin:
@frol, done.

AndrewPashkin:
@kvesteri, weird - I've got a `lint` fail on the file, that is not changed in this PR. I think it's because the linter is updated.

frol:
@kvesteri I am kindly pinging you as I am eager to clean my code using this helper util.

kvesteri:
GJ, now merged!

PR #254, opened 2016-11-08
Added SQLite support for PasswordType

frol:
This change fixes the issue of column type mismatch on SQLite as `VARBINARY` type fallbacks to `NUMERIC`, and in the end, Alembic always detects changes in my project:

```
Detected type change from NUMERIC(precision=128) to PasswordType(max_length=128) on 'user.password'
```

According to [this reply on SQLAlchemy support forum](https://groups.google.com/forum/#!topic/sqlalchemy/GpkSJ0XUgwo), `sa.VARBINARY` is not a good fallback, but I have no idea on how would a correct implementation look like, so I only added SQLite support.

Originally, I thought my problem is reported in #106, but it turned out to be another not quite closely related issue.

#233 is another enhancement for PasswordType, which, in fact, solves #106.

### DIFF
--- a/sqlalchemy_utils/types/password.py
+++ b/sqlalchemy_utils/types/password.py
@@ -2,7 +2,7 @@ import weakref
 
 import six
 from sqlalchemy import types
-from sqlalchemy.dialects import oracle, postgresql
+from sqlalchemy.dialects import oracle, postgresql, sqlite
 from sqlalchemy.ext.mutable import Mutable
 
 from ..exceptions import ImproperlyConfigured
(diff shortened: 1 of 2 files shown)


frol:
Weird! SQLite used NUMERIC type as a fallback for the column when it used to be VARBINARY, but once I tried to set it as NUMERIC on the Python side, it broke... Anyway, I didn't feel like NUMERIC is the proper type for PasswordType, so I choose `sqlite.BLOB` type for it.

@kvesteri All tests are green now.

kvesteri:
Thanks for the PR! :train:

PR #255, opened 2016-11-21
Allow calling CompositeType to create an instance of the backing comp…

rfw:
…osite type.

This allows you to do things like MyType(a=1, b=2) on instance creation.


kvesteri:
Could you add tests for this and some documentation as well (with preferably a real-life use case)?

kurtmckee:
PR #429 claims to replace this PR. As this PR hasn't been updated in five years, I recommend closing.

PR #257, opened 2016-11-23
Fixed an import error

vioan:
The import statement in an example was wrong. Now is fixed.


kvesteri:
Thanks for the PR! :train:

PR #264, opened 2017-03-14
Typo

DuncanBetts:
Very impressive! I think I will be using SQLAlchemy Utils extensively (I discovered it about an hour ago and have been reading through the docs). Thanks very much 👍.
I found a tiny typo.


kvesteri:
Thanks for the PR! :train:

PR #267, opened 2017-03-24
Fix error droping postgres database with beta version identifiers

scizo:
I ran into this trying to drop a database that was on a server
running version '9.4beta3'


kvesteri:
This should be fixed now (as of 0.32.14).

scizo:
Great. Thanks!

PR #277, opened 2017-06-29
Add more hints when decrypting AES with an invalid key

### LINKED ISSUE #275: UnicodeDecodeError in sqlalchemy_utils.types.encrypted
With 0.32.14 and python 3.4 we hit this on superset (https://github.com/airbnb/superset/issues/2600):
```
  File "/venv/lib/python3.4/site-packages/sqlalchemy_utils/types/encrypted.py", line 262, in process_result_value
    decrypted_value = self.engine.decrypt(value)
  File "/venv/lib/python3.4/site-packages/sqlalchemy_utils/types/encrypted.py", line 90, in decrypt
    decrypted = decrypted.decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf1 in position 2: invalid continuation byte
```

Any hint on what could be wrong?

xrmx:
Fix #275

### DIFF
--- a/sqlalchemy_utils/types/encrypted.py
+++ b/sqlalchemy_utils/types/encrypted.py
@@ -86,7 +86,10 @@ class AesEngine(EncryptionDecryptionBaseEngine):
         decrypted = decryptor.update(decrypted) + decryptor.finalize()
         decrypted = decrypted.rstrip(self.PADDING)
         if not isinstance(decrypted, six.string_types):
-            decrypted = decrypted.decode('utf-8')
+            try:
+                decrypted = decrypted.decode('utf-8')
+            except UnicodeDecodeError:
+                raise ValueError('Invalid decryption key')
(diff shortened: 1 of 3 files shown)


kvesteri:
Before merging this needs tests. Otherwise looks good :+1:

xrmx:
I've grepped decrypt in tests/ and haven't found anything. Where do you think should i add them?

kvesteri:
Each type has its own tests. You can find the encrypted type tests in https://github.com/kvesteri/sqlalchemy-utils/blob/master/tests/types/test_encrypted.py and add your tests there.

xrmx:
Updated PR with test plus removed an assert on an unrelated test that makes the build fails

xrmx:
Friendly ping

kvesteri:
Thanks for the PR! :train:

PR #280, opened 2017-08-21
Handle better the datetimes for EncryptedType

### LINKED ISSUE #184: EncryptedType with date time and timezone
Hi,

At https://github.com/kvesteri/sqlalchemy-utils/blob/master/sqlalchemy_utils/types/encrypted.py#L253, the date time is formatted with isoformat. Iso format convert the date time to str with microsecond and timezone.

At https://github.com/kvesteri/sqlalchemy-utils/blob/master/sqlalchemy_utils/types/encrypted.py#L278,
the formater does not take microsecond and timezone.

It is not really a bug, but the encrypted date time can not be other than a naive date time. I understand that convert all date time format is difficult, why does not use python-dateutils to convert str 2 date time, or regex.

konstantinoskostis:
This commit handles datetime and time objects that contain
microseconds and timezone information better, when decrypting
values that are subclasses of datetime.datetime, datetime.time
and datetime.date. The package 'python-dateutil' is required
if a user wants to use Datetimes with the EncryptedType.

Fixes #184

### DIFF
--- a/sqlalchemy_utils/types/encrypted.py
+++ b/sqlalchemy_utils/types/encrypted.py
@@ -21,6 +21,14 @@ except ImportError:
     pass
 
 
+dateutil = None
+try:
+    import dateutil
+    from dateutil.parser import parse as datetime_parse
+except ImportError:
+    pass
+
+
 class EncryptionDecryptionBaseEngine(object):
     """A base encryption and decryption engine.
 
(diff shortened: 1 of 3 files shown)


kvesteri:
We already have a bunch of optional dependencies. I would here introduce on optional dependency to python-dateutil, which is a mature package and much less error-prone than introducing our own regexp parsing.

konstantinoskostis:
@kvesteri hello! The PR has been already updated! 
IMHO if a user does not have (or does not want to use) the python-dateutil package then we should at least try to return  a `datetime.datetime` object accurate at least to the level of microseconds.

When you have the time, could you review the PR and make comments ? So that i can quickly make the necessary changes.

kvesteri:
I would like the implementation to be so that if user wishes to use this feature then it is required to use python-dateutil package. This is in line with other type implementations (for example ColorType which requires colour package).

konstantinoskostis:
PR is updated.

kvesteri:
Squash the commits, otherwise seems good to merge!

konstantinoskostis:
I will remove the `[TEMP]` flag.
Well, if this `assert isinstance(tz, pytz.tzfile.StaticTzInfo)` is kept in the test then Travis fails. See
```/opt/python/2.7.9/lib/python2.7/_strptime.py:317: ValueError
______________________ test_can_coerce_pytz_StaticTzInfo _______________________
    def test_can_coerce_pytz_StaticTzInfo():
        tzcol = TimezoneType(backend='pytz')
        tz = pytz.timezone('Pacific/Truk')
>       assert isinstance(tz, pytz.tzfile.StaticTzInfo)
E       AssertionError: assert False
E        +  where False = isinstance(<DstTzInfo 'Pacific/Truk' LMT+10:07:00 STD>, <class 'pytz.tzinfo.StaticTzInfo'>)
E        +    where <class 'pytz.tzinfo.StaticTzInfo'> = <module 'pytz.tzfile' from '/home/travis/build/kvesteri/sqlalchemy-utils/.tox/py27/lib/python2.7/site-packages/pytz/tzfile.pyc'>.StaticTzInfo
E        +      where <module 'pytz.tzfile' from '/home/travis/build/kvesteri/sqlalchemy-utils/.tox/py27/lib/python2.7/site-packages/pytz/tzfile.pyc'> = pytz.tzfile
tests/types/test_timezone.py:66: AssertionError ``` at https://travis-ci.org/kvesteri/sqlalchemy-utils/jobs/266900256

PR #282, opened 2017-09-20
Support for JSONType not native impl on PostgreSQL

KoNEW:
- Removed ImproperlyConfigured exception that will never be called cause of import json / anyjson in any case will have some value
- Added param native to JSONType declaration which will support to use TEXT datatype on database side instead of native JSON


kurtmckee:
This PR appears to have value but doesn't have unit tests hasn't been updated in five years. I recommend closing.

PR #283, opened 2017-09-28
Add Support for MSSQL uniqueidentifier type

nHurD:
This will enable utilization of MSSQL's native UUID type rather than falling back to BINARY(16) for persistence of UUIDs.


kvesteri:
Thanks for the PR!

PR #284, opened 2017-10-04
Align code with documentation

estyrke:
When using a url scheme of `postgres` (as in the docstrings), the wrong data base will be selected (`None`), and you get the not so helpful error message `FATAL:  database "<username>" does not exist`. Using the scheme `postgresql` works as intended. As far as I understand, both schemes work for sqlalchemy in general.


kvesteri:
Thanks for the PR! :train:

PR #286, opened 2017-10-05
Fix use of template0 and template1 with postgres.

funkybob:
template1 is the default template for creating a database with Postgres.
Also, tasks should connect to 'postgres' when issuing commands like CREATE DATABASE, not template1.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -7,4 +7,4 @@ prune docs/_build
 exclude docs/_themes/.git
 include tox.ini
 include *.rst
-exclude .git
+recursive-exclude .git/*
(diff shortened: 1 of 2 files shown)


kvesteri:
Is there a reason why this PR needs to modify MANIFEST.in?

funkybob:
Sorry, no.. was trying to fix some issues we were seeing with packaging, since we're using this branch until the PR is merged.

Am happy to undo that if you'd prefer.

funkybob:
Bump

kvesteri:
Merged manually, thanks for the PR! :train:

PR #287, opened 2017-10-05
Update aggregate columns when subclass objects are added to the relationship

fayazkhan:
Currently, `generator_registry` only looks for absolute class matches in the `session`. This would however skip over child classes of the relationship class.

Changed the behavior to use `isinstance` checks instead.


kvesteri:
Thanks for the PR! :train:

PR #293, opened 2017-10-12
updates existing translations

GrxE:
The content of the translation was not updated in the DB. Somehow SQLA was not aware of the changes made, no UPDATE statement was issued after `session.commit()`.


kvesteri:
Can you make test case that demonstrates the problem? Also I think it would be wise to use: http://docs.sqlalchemy.org/en/latest/orm/extensions/mutable.html

Issue #288, opened 2017-10-08
create_database doesn't close session


rmb938:
When creating multiple databases based on the same template pgsql errors with:

```
postgres_1  | ERROR:  source database "template1" is being accessed by other users
postgres_1  | DETAIL:  There is 1 other session using the database.
postgres_1  | STATEMENT:  CREATE DATABASE "right" ENCODING 'utf8' TEMPLATE template1
```

My python code:

```
create_database(uri_left, template="template1")
create_database(uri_right, template="template1")
```

Looking at the code it looks like the engine is never closed.

jforcada:
The other functions in the same module (drop_database, ...) do not close the connection either.

PR #295, opened 2017-11-09
Close connections on exists, create and drop database

Terseus:
The functions `database_exists`, `create_database` and `drop_database` all creates its own engines, however after using its connections it leaved them to the garbage collector.

This created some problems in some circumstances (e.g. calling twice `create_database` with different database names) because the connections were open when the second call is executed before the engine is garbage collected, see #288.

This try to fix it and add a regression test for PostgreSQL (unfortunately I don't know how to raise the error in MySQL or SQLite).

As a side change, fix the flake8 error E722, which was recently enabled by default in the latest flake8 versions.

### DIFF
--- a/sqlalchemy_utils/functions/database.py
+++ b/sqlalchemy_utils/functions/database.py
@@ -451,6 +451,13 @@ def database_exists(url):
 
     """
 
+    def get_result_scalar(engine, sql):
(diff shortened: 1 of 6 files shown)


kvesteri:
Thanks for the PR! :train: Merged manually

PR #296, opened 2017-11-09
Add optional JSON Schema and custom validator to JSONType

bollwyvl:
Hi folks! ❤️  the work!

This is a little PR for giving users the ability to validate their `JSONType`s with a [JSON Schema](http://json-schema.org/) as implemented by [jsonschema](https://github.com/Julian/jsonschema).

I initially started making it a separate type, but in my use of `JSONType`, I've basically never used it as a "random store of stuff I don't understand," but can go to that approach if desired.

An example use case: I'd like to store a Jupyter Notebook in a `JSONType`. I'd go pull the [nbformat schema](https://github.com/jupyter/nbformat/blob/master/nbformat/v4/nbformat.v4.schema.json), put it on a `notebook` column, and be able to confidently toss it into my database.

The custom `validator` stuff is a bit advanced, but is pretty important for being able to really hone in on how you want validation to work, especially if you are using existing schema that you don't control, or you want to use multi-file features.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,8 @@ extras_require = {
     'enum': ['enum34'] if sys.version_info < (3, 4) else [],
     'timezone': ['python-dateutil'],
     'url': ['furl >= 0.4.1'],
-    'encrypted': ['cryptography>=0.6']
+    'encrypted': ['cryptography>=0.6'],
+    'jsonschema': ['jsonschema>=2.6.0'],
 }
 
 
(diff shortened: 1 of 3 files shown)


bollwyvl:
I guess the linting on #297 will clear up the fails on this one...

bollwyvl:
Hooray lint!

kurtmckee:
This PR hasn't been updated in 4 years and has gone stale. I think validation of input data should happen at the application layer (whether using jsonschema, pydantic, or some other library), not the database layer.

I recommend closing this PR.

PR #297, opened 2017-11-09
Miscellaneous linting and test bugfixes

dargueta-work:
Master is broken and some of the tests mask errors, passing even if the tested code fails. This PR does the following:

* Fixed various bare `except` statements that the linter complained about.
* `test_invalid_phone_numbers` would always pass because of the bare try-except, even if the code it tests was broken.
* `test_invalid_phone_numbers_throw_dont_wrap_exception` would hide any exception thrown and fail with an `assert False` error, which is really hard to debug.
* Removed Python 2.6 from `tox`
* Added mention of support for Python 3.6 in the documentation.
* Altered installation instructions to include Python 3 steps as well.
* Add `flexmock` as a known third party library so `isort` doesn't complain about its placement

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,5 +1,6 @@
 [settings]
 known_first_party=sqlalchemy_utils
+known_third_party=flexmock
 line_length=79
 multi_line_output=3
 not_skip=__init__.py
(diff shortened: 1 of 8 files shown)


kvesteri:
Thanks for the PR! :train:

PR #305, opened 2018-01-30
Improve compatibility for migration tools.

9nix00:
thanks SQLAlchemy-Utils save my time.

there is some problem when work with flask-migrate. (may caused by alembic)

error like this:
```
File "... .../migrations/versions/c7bead11a48b_.py", line 39, in upgrade
    sa.Column('code', sqlalchemy_utils.types.uuid.UUIDType(length=16), nullable=False),
TypeError: __init__() got an unexpected keyword argument 'length'
```

that because UUIDType doesn't allow this argument. 
So I add `**kwargs` to improve  compatibility

### DIFF
--- a/sqlalchemy_utils/types/uuid.py
+++ b/sqlalchemy_utils/types/uuid.py
@@ -28,7 +28,7 @@ class UUIDType(types.TypeDecorator, ScalarCoercible):
 
     python_type = uuid.UUID
 
-    def __init__(self, binary=True, native=True):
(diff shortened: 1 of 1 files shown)


wangwenpei:
Does somebody can confirm this feature and merge it?

wangwenpei:
dear all,
I found this PR haven't merge into master. can you confirm it?
we have to use a hack version of sqlalchemy-utils, it is a little complex.

PR #308, opened 2018-02-04
fix #307: possibility to use othertype/discriminator than model class name

### LINKED ISSUE #307: generic_relationship with mapping on discriminator
I have a use for a generic_relationship on an existing schema but where the discriminator is not the name of the table but an integer than needs to be mapped to the name of a table.
e.g. 1 = "customers", 2 = "suppliers", 3 = "..."
Is it possible to specify such mapping ?
Something like: 
`generic_relationship(object_type, object_id, map_object_type_to_tablename)`
with 
`map_object_type_to_tablename = {1:"customers", 2:"vendors", ...}`

sdementen:


sdementen:
It aims to fix issue #307

sdementen:
it also fixes issue #309

PR #311, opened 2018-02-15
Add Ltree.descendant_of and Ltree.ancestor_of

kageurufu:
Matches behavior of equivalent LtreeType methods

Also add tests, using slice comparison

This allows for easier comparisons after query time, instead of needing to do
``model.ltreecol.path.startswith(other.ltreecol.path)`` or the like


kvesteri:
Thanks for the PR! :train:

PR #313, opened 2018-03-01
Fix emition of DDL statements.

daskol:
Type `Ltree` and similar are broken since it does not emits DDL at all on `GenericTypeCompiler`. It means that packages like `sadisplay` are not able to use ORM and to get type of columns. So, in this pull request I force `Ltree` to use `PGTypeCompiler` for expression compiling.


kvesteri:
Thanks for the PR. Can you add also unit tests for this?

PR #323, opened 2018-04-15
Set a value for typemap so references to Table.column.subfield work.

### LINKED ISSUE #319: CompositeType example doesn't work
Below is the example from pg_composite.py with a few extra pieces to fill in the gaps. On the pgsql side, I ran this, but I don't think the code gets that far to matter exactly how money_type is defined.
CREATE TYPE money_type AS (currency text, amount int);

It looks like Account is being replaced in the sqlalchemy.orm machinery and losing it's specialness. (Or I'm using it wrong.)

import sqlalchemy as sa
from sqlalchemy_utils import CompositeType, CurrencyType, register_composites
from sqlalchemy.ext.declarative import declarative_base
from sqlalchemy.orm import sessionmaker

_uri = ('postgresql+psycopg2://<snip>')

_engine = create_engine(_uri)
_conn = _engine.connect()

Base = declarative_base()
Session = sessionmaker()

class Account(Base):
    __tablename__ = 'account'
    balance = sa.Column(
        CompositeType(
            'money_type',
            [
                sa.Column('currency', CurrencyType),
                sa.Column('amount', sa.Integer)
            ]
        )
    )

register_composites(_conn)
session = Session()
session.query(Account).filter(Account.balance.amount > 5000)

  File "/usr/lib/python2.7/dist-packages/sqlalchemy/orm/attributes.py", line 193, in __getattr__
    key)
AttributeError: Neither 'InstrumentedAttribute' object nor 'Comparator' object associated with Account.balance has an attribute 'amount'

jo2y:
Fixes #319

### DIFF
--- a/sqlalchemy_utils/types/pg_composite.py
+++ b/sqlalchemy_utils/types/pg_composite.py
@@ -185,6 +185,7 @@ class CompositeType(UserDefinedType, SchemaType):
         SchemaType.__init__(self)
         self.name = name
         self.columns = columns
+        self.typemap = dict((col.name, col.type) for col in columns)
         if name in registered_composites:
             self.type_cls = registered_composites[name].type_cls
         else:


kvesteri:
Before merging this needs a test case.

orenleaffer-verkada:
It looks like there's a pull request with a unit test - https://github.com/kvesteri/sqlalchemy-utils/pull/323

Obviously this isn't a high priority, but I did just run into this issue. @kvesteri - to make this easier to accept, do you need a single PR with both the fix and the test?

orenleaffer-verkada:
@josecsotomorales tagging you since you appear to be actively maintaining this - the above question applies - I just ran into this issue and I believe that combined with https://github.com/StdCarrot/sqlalchemy-utils/commit/d7e0dc992179fa13c915915357164247a6f0dcc9 this fix both addresses the issue and has a unit test.

PR #333, opened 2018-07-16
fix(get_locale): set default locale to 'en' without Flask-Babel

### LINKED ISSUE #332: Cannot use Country name property without flask-babel
Hello, 

I have some issues while using the Country type from your library. Some simple property call will throw errors even though I only miss optional dependencies (or no dependency at all).

 See this sample code:
```
# In a new virtualenv
$ pip install -e .
$ python
Python 2.7 (r27:82500, May 28 2018, 14:40:35)
>>> from sqlalchemy_utils import Country
>>> Country("FR").name
  File "sqlalchemy_utils/primitives/country.py", line 57, in __init__
    self.validate(code_or_country)
  File "sqlalchemy_utils/primitives/country.py", line 74, in validate
    i18n.babel.Locale('en').territories[code]
AttributeError: 'NoneType' object has no attribute 'Locale'

# [...]

$ pip install -e .[babel]
$ python
Python 2.7 (r27:82500, May 28 2018, 14:40:35)
>>> from sqlalchemy_utils import Country
>>> Country("FR").name
  File "sqlalchemy_utils/primitives/country.py", line 69, in name
    return i18n.get_locale().territories[self.code]
  File "sqlalchemy_utils/i18n.py", line 20, in get_locale
    'Could not load get_locale function from Flask-Babel. Either '
sqlalchemy_utils.exceptions.ImproperlyConfigured: Could not load get_locale function from Flask-Babel. Either install Flask-Babel or make a similar function and override it in this module.
```

In the first case, since you can't validate your Country without babel, I suggest it could become a required dependency, or deactivate the validation if `babel is None`.

In the second case, I do not necessarily want to rely on `flask-babel` to retrieve my locale through a request:
https://github.com/kvesteri/sqlalchemy-utils/blob/abf6e65f8a2f7ad98326b5cc2e152b8584230392/sqlalchemy_utils/i18n.py#L15-L23

 So it could be great to add a _working_ fallback behaviour, as you already do in some tests: 
https://github.com/kvesteri/sqlalchemy-utils/blob/abf6e65f8a2f7ad98326b5cc2e152b8584230392/tests/types/test_currency.py#L9

If you agree with those changes, I can provide you a PR fixing those issues. What do you think about it ?

aveuiller:
Closes #332

### DIFF
--- a/sqlalchemy_utils/i18n.py
+++ b/sqlalchemy_utils/i18n.py
@@ -12,13 +12,15 @@ try:
 except ImportError:
     babel = None
 
+
 try:
-    from flask_babel import get_locale
-except ImportError:
+    def get_locale(): return babel.Locale('en')
+except AttributeError:
+    # As babel is optional, we may raise an AttributeError accessing it
     def get_locale():
         raise ImproperlyConfigured(
-            'Could not load get_locale function from Flask-Babel. Either '
-            'install Flask-Babel or make a similar function and override it '
+            'Could not load get_locale function using Babel. Either '
+            'install Babel or make a similar function and override it '
             'in this module.'
         )
 
--- a/sqlalchemy_utils/primitives/country.py
+++ b/sqlalchemy_utils/primitives/country.py
@@ -25,7 +25,8 @@ class Country(object):
 
         Country(Country('FI')).code  # 'FI'
 
-    Country always validates the given code.
+    Country always validates the given code if you use at least the optional
+    dependency list 'babel', otherwise no validation are performed.
 
     ::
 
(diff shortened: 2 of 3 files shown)


sqlalchemy_utils/i18n.py
```diff
@@ -12,13 +12,15 @@
...
+
 try:
-    from flask_babel import get_locale
-except ImportError:
+    def get_locale(): return babel.Locale('en')
+except AttributeError:
+    # As babel is optional, we may raise an AttributeError accessing it
     def get_locale():
         raise ImproperlyConfigured(
             'Could not load get_locale function from Flask-Babel. Either '
```
kdeldycke: Now that `Flask-Babel` is no longer required I guess you can remove mention of it in the error message.

kdeldycke:
Any news on that one? :)

kvesteri:
Thanks for the PR! :train:

kdeldycke:
Thanks @kvesteri for the merge and release! 👍

PR #334, opened 2018-07-30
Removed line from get_referencing_foreign_keys() which skips source t…

shatsky:
…able when building referencing_foreign_keys set; it caused skipping foreign key fields in source table referencing its own fields, which can exist if table represents tree structure

### DIFF
--- a/sqlalchemy_utils/functions/foreign_keys.py
+++ b/sqlalchemy_utils/functions/foreign_keys.py
@@ -88,12 +88,11 @@ def get_referencing_foreign_keys(mixed):
     referencing_foreign_keys = set()
 
     for table in mixed.metadata.tables.values():
-        if table not in tables:
-            for constraint in table.constraints:
-                if isinstance(constraint, sa.sql.schema.ForeignKeyConstraint):
-                    for fk in constraint.elements:
-                        if any(fk.references(t) for t in tables):
-                            referencing_foreign_keys.add(fk)
+        for constraint in table.constraints:
+            if isinstance(constraint, sa.sql.schema.ForeignKeyConstraint):
+                for fk in constraint.elements:
+                    if any(fk.references(t) for t in tables):
+                        referencing_foreign_keys.add(fk)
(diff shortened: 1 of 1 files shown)


shatsky:
I see the test is failing because Admin declarative class inherits from User and `get_tables(Admin)` returns not only the table specified in the Admin, but also the table specified in the User. I wonder why `sa.orm.class_mapper(DeclarativeWithASingleTableInheritedFromParentWithItsOwnTable).tables` which get_tables() relies upon contains both tables. Docs state that "If the mapper is mapped to a Join, or an Alias representing a Select, the individual Table objects that comprise the full construct will be represented here." - but this isn't the case.
Is patching get_tables() to return `[mixed.__table__]` if mixed is `sa.ext.declarative.api.DeclarativeMeta` a proper solution?

kvesteri:
I think it is the right way that ``get_tables()`` returns all tables for polymorphic declarative classes, hence I think in this case we should just change the behaviour and the test to return also the foreign key that points to user table.

Also I would like this PR to contain one additional test for the tree relationship.

shatsky:
Now I see I've missed joined table inheritance.

> we should just change the behaviour and the test to return also the foreign key that points to user table

I can't agree, if get_referencing_foreign_keys() purpose is to find fks that can point to the given object from other [dependent] objects, we don't want fks which represent relation inside the object. So with polymorphic model with joined table inheritance I guess we can consider that fk in one of its tables referencing another one is always connecting two rows which make dataset of a single model instance and has to be skipped:
```
    for source_table in mixed.metadata.tables.values():
        for constraint in source_table.constraints:
            if isinstance(constraint, sa.sql.schema.ForeignKeyConstraint):
                for fk in constraint.elements:
                    for target_table in tables:
                        if fk.references(target_table) and (source_table not in tables or source_table is target_table):
                            referencing_foreign_keys.add(fk)
                            break
```

kvesteri:
True, I agree with you :+1:

PR #337, opened 2018-08-03
Add support for MSSQL when creating/dropping db

jomasti:


jomasti:
I tested this locally using a SQL Server instance in Docker. I guess the same would need to be done in Travis to support running tests against SQL Server.

deterralba:
I wanted to ask if we the support for sql server would be welcome for the create/drop operation, I am glad to see it's an ongoing work!

The value of `engine.url.database` is None when we use exact pyobdc strings (see http://docs.sqlalchemy.org/en/latest/dialects/mssql.html#pass-through-exact-pyodbc-string), hence the database creation fails.
Should this be patched here or should I create an issue the sqlalchemy repo?

kvesteri:
Thanks for the PR. In order for this to get merged the travis CI must be configured to support MSSQL as well.

jomasti:
@kvesteri Travis is set up with SQL Server now.

kvesteri:
Thanks for the PR! :train:

Issue #347, opened 2018-10-31
Doesn't support shortcodes in PhoneNumberType


anandtripathi5:
PhoneNumberType of column in SQLalchemy models doesn't support short-codes and toll-free.

anandtripathi5:
@kvesteri review it and be active please.

PR #348, opened 2018-10-31
#347 support of short-code in phone number type

anandtripathi5:
Use of _check_region parameter of phonenumbers library. If this parameter's value is false then it will not check for the region it simple bypasses the region


kvesteri:
Can you please add one unit test for this that showcases the intended use?

anandtripathi5:
@kvesteri its done, please have a look

anandtripathi5:
@kvesteri hi please check the test cases are written already and merge the request cause its difficult for me cause it is not accepting the shortcodes
Thanks

PR #356, opened 2019-01-10
Fix #303: Support creating and dropping a PostgreSQL database when using pg8000 driver

### LINKED ISSUE #303: pg8000 support for create_database
`create_database` has a check for `engine.driver == 'psycopg2'`, which it then uses to enable autocommit for the engine, which (AFAIK) avoids an error about trying to issue CREATE DATABASE from within a transaction. Something similar (see https://github.com/zzzeek/sqlalchemy/pull/88) needs to be done for pg8000 to avoid the same issue.

https://github.com/kvesteri/sqlalchemy-utils/blob/master/sqlalchemy_utils/functions/database.py#L540

mohaseeb:
As reported by #303 current implementation of create_database/drop_database doesn't support creating/dropping a PostgreSQL database when connecting using [pg8000 driver](https://docs.sqlalchemy.org/en/latest/dialects/postgresql.html#module-sqlalchemy.dialects.postgresql.pg8000). The reason is when pg8000 driver is used (i.e. when the drivername is `postgresql+pg8000`), create_database/drop_database doesn't set the isolation level needed to execute the 'create database'/'drop database' command on PostgreSQL. This PR is fixing that.
- Commits  5dec6f23883bc411a4e9136db781a6de1e442239 and ebd6b10b8aee827e1f84202e397d2f3a365b0a77 add a TC that tests creating a PostgreSQL DB using pg8000. Only the TC without the fix was added first so the travis-ci job fails to reproduce the error ([see the TC failure here](https://travis-ci.org/kvesteri/sqlalchemy-utils/jobs/478001813#L3506)).
- The following commit adds support to create a PostgreSQL DB using pg8000.
- Commit 49ba768ed4d8d1b57acca6f5ffef8ecba13cc9fc extends the tests to test dropping a PostgreSQL DB using pg8000. The TC fails as intended to reproduce the error as can be seen [here](https://travis-ci.org/kvesteri/sqlalchemy-utils/jobs/478651095#L3302).
- The following commit adds support to drop a PostgreSQL DB using pg8000.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,7 @@ extras_require = {
         'flexmock>=0.9.7',
         'mock==2.0.0',
         'psycopg2>=2.5.1',
+        'pg8000>=1.12.4',
         'pytz>=2014.2',
         'python-dateutil>=2.2',
         'pymysql',
--- a/sqlalchemy_utils/__init__.py
+++ b/sqlalchemy_utils/__init__.py
@@ -100,4 +100,4 @@ from .view import (  # noqa
     refresh_materialized_view
 )
 
-__version__ = '0.33.10'
+__version__ = '0.33.11.dev1'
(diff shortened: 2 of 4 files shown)


mohaseeb:
@kvesteri this PR is ready for review, please have a look when possible. Thanks!

kvesteri:
Thanks for the PR! :train:

Issue #360, opened 2019-01-29
Should Country be orderable?


TrilceAC:
Suppose that your models must support Persons which might have many Nationalities and you decide to proceed using a `Person` class and another `Nationality` class to represent the nationalities of persons by means of a foreign key column to the person id and a CountryType column to represent the related nacionality:

```
class Persona(Model):
    id = Column(Integer, primary_key=True, index=True)
    name = Column(Unicode(35), nullable=False, index=True)
    surname = Column(Unicode(35), nullable=False, index=True)

class Nationality(Model):
    person_id = Column(Integer,
                        ForeignKey('person.id',
                                   ondelete='CASCADE'),
                        primary_key=True,
                        index=True)

    person = relationship(
        'Person',
        backref=backref(
            'nationalities',
            cascade='all, delete-orphan'
        )
    )

    country = Column(CountryType, primary_key=True, index=True)

```
No orphan is desired, therefore `ondelete='CASCADE'` is used on the foreign key, as well as `cascade='all, delete-orphan'` is used on the relationship.

Suppose that john is added and commited:

```
>>>john = Person(name='John', surname='Doe', nationalities=[Nationality(country='ES'), Nationality(country='EU')])
>>>db.session.add(john)
>>>db.session.commit()

```
At this point, if one tries to delete `john`, the behaviour would depend on whether the object is loaded on the session or not. If the object is not loaded, the deletion will succeed, because the RDBMS is in charge of deleting the related rows and it knows how to do that:

```
>>>db.session.delete(john)
>>>db.session.commit()

```
But if the object is loaded in the session, a `TypeError: unorderable types: Country() < Country()` exception raises:

```
>>># Remember to recreate john if you deleted it!!!
>>>john
<Person(3, 'John Doe')>
>>># Calling john loads it from the database.
>>> db.session.delete(john)
>>> db.session.commit()

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".virtualenvs/gesific/lib/python3.5/site-packages/sqlalchemy/orm/scoping.py", line 153, in do
    return getattr(self.registry(), name)(*args, **kwargs)
  File ".virtualenvs/gesific/lib/python3.5/site-packages/sqlalchemy/orm/session.py", line 943, in commit
    self.transaction.commit()
  File ".virtualenvs/gesific/lib/python3.5/site-packages/sqlalchemy/orm/session.py", line 467, in commit
    self._prepare_impl()
[...]

PR #361, opened 2019-01-29
Country implements ordering

TrilceAC:
`Country` implements a total ordering based on `Country.code`. This ordering is also comparable to any string. This solves #360

### DIFF
--- a/sqlalchemy_utils/primitives/country.py
+++ b/sqlalchemy_utils/primitives/country.py
@@ -95,6 +95,12 @@ class Country(object):
     def __ne__(self, other):
         return not (self == other)
 
+    def __lt__(self, other):
+        if isinstance(other, Country):
+            return self.code < other.code
+        if isinstance(other, str):
+            return self.code < other
+
     def __repr__(self):
         return '%s(%r)' % (self.__class__.__name__, self.code)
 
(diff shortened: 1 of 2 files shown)


kvesteri:
This needs at least couple of unit tests. Also why not use total_ordering decorator from functools module?

TrilceAC:
I'll try to prepare some tests.

Regarding the use of tolta_ordering, if you want, I can use it. I did not just because of the note at total_ordering, and because it was easy to implemente the four methods

> Note: While this decorator makes it easy to create well behaved totally ordered types, it does come at the cost of slower execution and more complex stack traces for the derived comparison methods. If performance benchmarking indicates this is a bottleneck for a given application, implementing all six rich comparison methods instead is likely to provide an easy speed boost.

Probably performance will never be an issue, but every little bit helps.

kvesteri:
I'm all for code clarity and simplicity over micro-optimizations, hence please use total_ordering even if there is a small performance penalty (as there is with any decorators).

For tests you can just have single test method [parametrized](https://docs.pytest.org/en/latest/parametrize.html)

```python
@pytest.mark.parametrize(
   'assertion',
   [
       Country('FI') > Country('ES'),
       ...
   ]
)
def test_ordering(assertion):
    assert assertion
```

TrilceAC:
I have implemented the changes that you propose. Please check.

PR #362, opened 2019-02-05
STAN-79 Add support for Oracle RDBMS (not for Oracle 12)

uudecode:
I add little Oracle support. Possible, will after more after. Thank you.


kvesteri:
If we are going to introduce Oracle support I think all the database functions should then support Oracle. There should also be test coverage for Oracle and those tests would need to be run in Github CI. Until these conditions are met, I don't think we should add Oracle support for individual functions - thus closing this.

Issue #330, opened 2018-06-06
database_exists() throws exception when PostgreSQL role does not have access to postgres database


williamjacksn:
My PostgreSQL role has access to my personal database but not the `postgres` database. When I call `database_exists()`, I get an exception:

(passwords, hostnames, and IP addresses obscured)

```python
>>> sqlalchemy_utils.database_exists('postgres://william:secretpassword@db.example.com:5433/william')
Traceback (most recent call last):
  File "/home/william/.pyenv/versions/3.6.5/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 2158, in _wrap_pool_connect
    return fn()
  File "/home/william/.pyenv/versions/3.6.5/lib/python3.6/site-packages/sqlalchemy/pool.py", line 403, in connect
    return _ConnectionFairy._checkout(self)
  File "/home/william/.pyenv/versions/3.6.5/lib/python3.6/site-packages/sqlalchemy/pool.py", line 791, in _checkout
    fairy = _ConnectionRecord.checkout(pool)
  File "/home/william/.pyenv/versions/3.6.5/lib/python3.6/site-packages/sqlalchemy/pool.py", line 532, in checkout
    rec = pool._do_get()
  File "/home/william/.pyenv/versions/3.6.5/lib/python3.6/site-packages/sqlalchemy/pool.py", line 1196, in _do_get
    self._dec_overflow()
  File "/home/william/.pyenv/versions/3.6.5/lib/python3.6/site-packages/sqlalchemy/util/langhelpers.py", line 66, in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
  File "/home/william/.pyenv/versions/3.6.5/lib/python3.6/site-packages/sqlalchemy/util/compat.py", line 187, in reraise
    raise value
  File "/home/william/.pyenv/versions/3.6.5/lib/python3.6/site-packages/sqlalchemy/pool.py", line 1193, in _do_get
    return self._create_connection()
  File "/home/william/.pyenv/versions/3.6.5/lib/python3.6/site-packages/sqlalchemy/pool.py", line 350, in _create_connection
    return _ConnectionRecord(self)
  File "/home/william/.pyenv/versions/3.6.5/lib/python3.6/site-packages/sqlalchemy/pool.py", line 477, in __init__
    self.__connect(first_connect_check=True)
[...]

PR #372, opened 2019-05-13
database_exists: don't connect to 'postgres' data base for database existence check

bernt-matthias:
On our installation we don't have CONNECT permission for the data base called 'postgres'. Hence the function did not work. 

This has been introduced here (the if else branch probably made sense in drop and create, but it doesn't here in my opinion)

https://github.com/kvesteri/sqlalchemy-utils/commit/5d741c16662cde063e6930e7d3fe87b00ae9b8bb
https://github.com/kvesteri/sqlalchemy-utils/commit/d7f2905a39a02a2aa5b7b81640abe8cfd07e8771

### DIFF
--- a/sqlalchemy_utils/functions/database.py
+++ b/sqlalchemy_utils/functions/database.py
@@ -469,11 +469,6 @@ def database_exists(url):
 
     url = copy(make_url(url))
     database = url.database
-    if url.drivername.startswith('postgres'):
-        url.database = 'postgres'
-    else:
-        url.database = None
-
     engine = sa.create_engine(url)
 
     if engine.dialect.name == 'postgresql':
--- a/sqlalchemy_utils/functions/database.py
+++ b/sqlalchemy_utils/functions/database.py
@@ -468,7 +468,7 @@ def database_exists(url):
         return header[:16] == b'SQLite format 3\x00'
 
     url = copy(make_url(url))
-    database = url.database
+    database, url.database = url.database, None
     engine = sa.create_engine(url)
 
     if engine.dialect.name == 'postgresql':


bernt-matthias:
ping @mvdbeek @erasche

bernt-matthias:
Maybe the same should be done for create and drop data base?

erseco:
This PR will be really helpfull when running on Heroku, because they remove **CONNECT** privilege in `postgres` database

You can replicate this behavior in your local database:

```
CREATE DATABASE test;
CREATE USER testuser WITH PASSWORD 'testpassword';
REVOKE CONNECT ON DATABASE postgres FROM PUBLIC;
REVOKE ALL PRIVILEGES ON DATABASE postgres from testuser;
```

Now you can connect using `DATABASE_URL='postgres://testuser:testpassword@localhost/test'` and get the **CONNECT** error when calling to `database_exists()` function.

sveinugu:
@kvesteri A gentle nudge in your direction in order to get this pull request accepted.

I am in the exact same same situation as the issue creator @bernt-matthias, that I cannot directly deploy a production instance of the Galaxy framework for life science research (https://github.com/galaxyproject/galaxy), as our database is hosted on an university-wide service that does not provide `CONNECT` access to the `postgres` database. 

There has been a need to implement workarounds downstream for a longer time (https://github.com/galaxyproject/galaxy/issues/7941, https://github.com/usegalaxy-eu/galaxy/commit/59305effd73c0b6c195afa9cb0f4f0d1ca8e994f, https://github.com/CTFd/CTFd/pull/1403#discussion_r426710601), and I would rather not have to implement another one. 

It seems to me that this pull request should not be controversial.

sveinugu:
Should also fix https://github.com/kvesteri/sqlalchemy-utils/issues/330.

CaselIT:
This change breaks `database_exists` if an user does not have a db called like username.

It errors with `(psycopg2.OperationalError) FATAL:  database "<username-here>" does not exist`

CaselIT:
It is probably better to use one of the template database

bernt-matthias:
Yep, I can reproduce this from the logs in Galaxy's testing framework (https://github.com/galaxyproject/galaxy/pull/9967). Your suggestion is to revert the `if .. else ..` block but use a different value for `url.database`? Any suggestions?

CaselIT:
Template0 or template1 should probably work. But it was reverted in this commit https://github.com/kvesteri/sqlalchemy-utils/commit/d7f2905a39a02a2aa5b7b81640abe8cfd07e8771 so they may have problems.

Probably the best solution is to add a parameter in the function, called maybe `postgresql_db=None` that can be used to specify an existing db by an user.

bernt-matthias:
did some experiments (for postgres .. since the PR should not change anything for other DBs).  The following works for me (python 3.6, sqlalchemy 1.3.17, psycopg2-binary 2.8.4):

```
from copy import copy
import sqlalchemy as sa
from sqlalchemy.engine.url import make_url

def ex(url):
    url = copy(make_url(url))
    engine=sa.create_engine(url)
    result_proxy = engine.execute("SELECT 1 FROM pg_database WHERE datname='%s'" % database)
    return result_proxy.scalar()

print(ex("postgresql://USER:PASSWORD@HOST/DATABASE"))
```

works. So even if I see the error in the Galaxy CI logs .. I can not reproduce it.

Any idea where the information on the user may be originating from in the error case?

CaselIT:
by default postgres connects to a db called as the user, so if a db with that name exits the check has success, otherwise it errors.

To reproduce just create a new user in the db and connect with that user.

bernt-matthias:
Do you think that there is always a "public" data base, i.e. one with a connect privilege?

CaselIT:
From the [documentation](https://www.postgresql.org/docs/12/manage-ag-templatedbs.html) `template0` and `template1` should always be defined, (unless someone has not removed them) so I think we could use `template0` for the connection (since `template1` is the default db used then creating a db). 
But I think the best solution is to add an argument to the function and set it to `postgres` or `template0` as default, so that it can be change in case of unusual situations.

bernt-matthias:
Was thinking more about a good solution. How about having an additional parameter `postgres_db=["template0", "template1", "postgres"]` and we check all entries in the list. If one does not raise an exception its fine, otherwise the user can another value. 

In this way the function can be used in a wider variety of deployments without configuration.

CaselIT:
yes, that also works

bernt-matthias:
Some ideas over here https://github.com/kvesteri/sqlalchemy-utils/pull/463

PR #376, opened 2019-06-17
Fix generic_repr with eq comparison.

gnattishness:
In some cases, it is possible for `value == NO_VALUE` to raise an exception (when `value` is a class that defines `__eq__()`)

As `NO_VALUE` is a singleton, it is fine to check using `is`, which will never raise exceptions.


gnattishness:
@kvesteri looks like this build error is expected due to 3.4 deprecation?

kvesteri:
Before merging this needs a test case. Could you add one?

gnattishness:
Sorry for the delay. The added test is a bit convoluted but correctly demonstrates the the potential problem that was fixed.

Using `==` in this case is unnecessary and can cause side-effects. (Though I would think it fair to say that a class whose `__eq__()` can raise exceptions is buggy and should be fixed to return `NotImplemented`. )

Please let me know if I can assist further and thanks for your work!

gnattishness:
@kvesteri Forgot to check linting, should be all good.

PR #377, opened 2019-06-19
Enable JSONB function

getglad:
Small modification to enable the use of `jsonb` in addition to `json` for Postgres operations


kvesteri:
Rather than introducing a parameter like this I'd like to see a function named ``jsonb_sql`` which would then return jsonb structs.

getglad:
@kvesteri made those changes - let me know what you think

PR #378, opened 2019-06-20
Get class by Table Name

getglad:
Method to extract table out of base by a given table name


kvesteri:
Closing this as we already have a function called `get_class_by_table`. I'd like to hear the rationale for this.

PR #379, opened 2019-06-26
Fix deprecation warnings

Le-Stagiaire:
Fix all deprecation warnings coming from sqlalchemy-utils, there still one coming from sqlalchemy itself though


Le-Stagiaire:
This fails not because of the code I changed I think, only on python3.4

PR #381, opened 2019-07-16
Add support date and datetime for PendulumType

yk-lab:
support for [pendulum](https://pendulum.eustace.io/).

https://github.com/sdispater/pendulum/issues/155


kvesteri:
Instead of just adding this I'd like to have some kind of combination of this and ArrowType and any other easy to use datetime library one might want to add support for in the future. Something like EnrichedDateTime (naming suggestions welcome!). And then the one could just select the library used for that.

yk-lab:
OK. Just a moment please.

yk-lab:
I thought it was a good idea and rewritten code.

ulyssesxxxi:
Any idea on when this will be ready? In my case it is very much needed. Thanks a lot in advance.

PR #384, opened 2019-07-25
add hash method for PhoneNumber

segfault87:
`PhoneNumberType` does not work with current SQLAlchemy because it tries to hash `PhoneNumber` when querying.

This PR implements hash function for PhoneNumber class.


kvesteri:
Please add a unit test for this as well.

PR #388, opened 2019-08-01
Validate that value is a valid choice.

philip-newzoo:
We should validate when loading from JSON too.


kvesteri:
Please add a test case for this.

CodePint:
Hello, I am still running into this bug. 

Im happy to pick this up when I have a bit of time

CodePint:
test file is here. I'll come back and update/fork.
https://github.com/kvesteri/sqlalchemy-utils/blob/master/tests/types/test_choice.py

PR #389, opened 2019-08-01
Update database.py

jssuzanne:
The create_database and drop_database with 'mssql+pymssql' must also do in autocommit mode


GohuHQ:
Hi,
Yes i'm expecting the same issue and this fixes the problem.

Thanks !

GohuHQ:
It seems like there is a problem with the travis...
Maybe we just need to reexecute the build ?

PR #391, opened 2019-08-02
Update password.py

jssuzanne:
MsSQL can't convert the result of the encrypt to VARBINARY
```
 b'Implicit conversion from data type varchar to varbinary is not allowed. Use the CONVERT function to run this query.DB-Lib error message 20018, severity 16:\nGeneral SQL Server error: Check messages from the SQL Server\n')
```


kurtmckee:
@jssuzanne Thanks for this work! Right now there's no setup to run tests against MSSQL, unfortunately.

@kvesteri I recommend closing this PR simply because there's no way to run tests against MSSQL. It will be good to set that up in CI so contributors' pull requests can benefit from that setup.

PR #403, opened 2019-11-03
Add support date and datetime for pendulum

yk-lab:
Update #381 .

Coding new type that EnrichedDateType and EnrichedDateTimeType.
There are supported for arrow and pendulum.


sqlalchemy_utils/types/enriched_datetime/enriched_datetime_type.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+        user = User()
+        user.created_at = pendulum.now()
+        session.add(user)
+        session.commit()
+    """
+    impl = types.DateTime
+
+    def __init__(self, type="pendulum", *args, **kwargs):
```
kvesteri: I think a better way would be to pass the ``date_class`` as an argument.

```
def __init__(self, date_class= PendulumDate, *args, **kwargs):
```

kvesteri: The constructor of those custom date classes could throw an error if the given package is not installed.

sqlalchemy_utils/types/enriched_datetime/pendulum_datetime.py
```diff
@@ -0,0 +1,38 @@
...
+except ImportError:
+    pass
+
+
+class PendulumDatetime(object):
+    def _coerce(self, impl, value):
+        if value is not None:
+            if isinstance(value, pendulum.DateTime):
+                pass
+            elif isinstance(value, int):
```
wakemaster39: Sorry to jump in randomly I was looking for this and was testing it out. This should also support floats

`isinstance(value, (int, float))`

PR #405, opened 2019-11-04
Fix base class in generic relationship doc

merwok:
This is a draft PR because I haven’t searched the rest of the docs for the same issue.  Github search is surprisingly very bad for searching an exact code fragment!


merwok:
Found the same issue in docstrings in
sqlalchemy_utils/types/phone_number.py
sqlalchemy_utils/functions/foreign_keys.py

PR #406, opened 2019-11-26
Add support to specify CASCADE when dropping views

amicks:
I ran into an issue with SQLite where views would be dropped, resulting in a SQLAlchemy.exc.OperationalError for syntax on the "CASCADE" suffix.

Some SQL DBMSs like MSSQL or SQLite don't support the CASCADE suffix so the caller should be able to specify this.  Otherwise, dropping views is not possible on those DBMSs.

### DIFF
--- a/sqlalchemy_utils/view.py
+++ b/sqlalchemy_utils/view.py
@@ -20,16 +20,18 @@ def compile_create_materialized_view(element, compiler, **kw):
 
 
 class DropView(DDLElement):
-    def __init__(self, name, materialized=False):
+    def __init__(self, name, materialized=False, cascade=True):
         self.name = name
         self.materialized = materialized
+        self.cascade = cascade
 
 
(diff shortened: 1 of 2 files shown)


sqlalchemy_utils/view.py
```diff
@@ -149,7 +152,11 @@ def create_indexes(target, connection, **kw):
         for idx in table.indexes:
             idx.create(connection)
 
-    sa.event.listen(metadata, 'before_drop', DropView(name))
+    sa.event.listen(
```
amicks: This change isn't functional and is only because CI was failing on the lint job for too many characters in one row.

amicks:
CI build is failing on unrelated code in encryption TestAesGcmEngine.

amicks:
Build is passing now :)

kvesteri:
Thanks for the PR! 🚋

PR #408, opened 2019-12-08
Remove explain and explain_analyze

kvesteri:


swirlingsand:
Hey, maybe this could be removed from docs?
https://sqlalchemy-utils.readthedocs.io/en/stable/database_helpers.html#analyze

image

 ```
AttributeError: module 'sqlalchemy_utils.functions' has no attribute 'analyze'
```
(makes sense seeing this commit)

lpsinger:
Why was this removed?

kvesteri:
SQLAlchemy internals (which this function relied on) changed a lot and thus the function stopped working. It was never a good idea to use those internals in the first place.

PR #415, opened 2020-01-23
A couple bugfixes and a feature I needed (self-referencing tables)

ericatkin:


ericatkin:
So, there are 6 failing tests, but the failures existed before my changes. I haven't looked at them. I have added tests for my changes and they are passing.

PR #424, opened 2020-03-12
Fix UUIDType repr

ziima:
Fix `__repr__` for `UUIDType`, mainly to help alembic render correct output.


kvesteri:
Thanks for the PR! 🚋

PR #426, opened 2020-03-15
Using String instead of LargeBinary for impl of EncryptedType

### LINKED ISSUE #425: EnryptedType in python3 and postgres fails with 'string argument without an encoding'
When querying a record with a column of EncryptedType in python3.7 with a MySQL backend and packages

- SQLAlchemy==1.3.15
- SQLAlchemy-Utils==0.36.1
- psycopg2-binary==2.8.4

The query fails with TypeError: 'string argument without an encoding'.

The issue appears to be that EncryptedType has an impl of LargeBinary which in python3 processes the returned value with the command

This issue is a duplicate of https://github.com/kvesteri/sqlalchemy-utils/issues/366. It appears that the original author of #366 closed the issue because they resolved their issue in the short term by using a different driver, but this issue is reproducible in `psycopg2-binary==2.8.4` which is a popular choice for postgres.

This is happening because a string gets converted to bytes in python3 without specifying the encoding inside sqlalchemy's `_Bytes` type (base of `LargeBytes`)

https://github.com/sqlalchemy/sqlalchemy/blob/7c6bdc9f0d279be40d6f53b68a3a448b2ccfb73f/lib/sqlalchemy/sql/sqltypes.py#L929

```python
def process(value):
    if value is not None:
        value = bytes(value)
    return value
```

If we manually fix that issue in sqlalchemy, we run into another problem

```
  File "/Users/aicioara/dev/python/motuz/venv/lib/python3.7/site-packages/sqlalchemy_utils/types/encrypted/encrypted_type.py", line 119, in decrypt
    decrypted = base64.b64decode(value)
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/base64.py", line 87, in b64decode
    return binascii.a2b_base64(s)
binascii.Error: Invalid base64-encoded string: number of data characters (49) cannot be 1 more than a multiple of 4
```

It looks like an encrypted key such as

```
u4l3U4adiXfWV0vbD5C6kQ==
```

Gets transformed into

```
<psycopg2.extensions.Binary object at 0x10c9d7a80>

# where
# .__str__() == "'u4l3U4adiXfWV0vbD5C6kQ=='::bytea"
```

And gets saved into the database as

```
\x75346c33553461646958665756307662443543366b513d3d
```

Which looks like it's some base16 encoding of the ASCII value of `u4l3U4adiXfWV0vbD5C6kQ==`

aicioara:
Closes #425

LargeBinary is problematic in python3 (see issue description). I recommend that we use `String` instead of `LargeBinary` as the impl for `EncryptedType`.

`self.fernet.encrypt()` and all other `encrypt()` methods in this class already return base64 encodings of the data, they are just formatted as `byte` instead of `str`. It does not make much sense to store them as binary because:

- encoding/decoding is problematic
- They are BASE64 (i.e. string) already

I manually tested this change by encoding and decoding strings using all three encryption engines (`AesEngine`, `AesGcmEngine`, `FernetEngine`) and everything works as expected. 

Additionally, I went ahead and included 

```
git+git://github.com/aicioara-forks/sqlalchemy-utils@0.36.1-fix-encrypted-type#egg=SQLAlchemy-Utils
```

In my `requirements.txt` and my entire flask application works as expected.

Happy to take feedback to make this perfect.

### DIFF
--- a/sqlalchemy_utils/types/encrypted/encrypted_type.py
+++ b/sqlalchemy_utils/types/encrypted/encrypted_type.py
@@ -110,7 +110,7 @@ class AesEngine(EncryptionDecryptionBaseEngine):
         encryptor = self.cipher.encryptor()
         encrypted = encryptor.update(value) + encryptor.finalize()
         encrypted = base64.b64encode(encrypted)
-        return encrypted
+        return encrypted.decode('utf-8')
 
     def decrypt(self, value):
         if isinstance(value, six.text_type):
@@ -164,7 +164,7 @@ class AesGcmEngine(EncryptionDecryptionBaseEngine):
         encrypted = encryptor.update(value) + encryptor.finalize()
         assert len(encryptor.tag) == self.TAG_SIZE_BYTES
         encrypted = base64.b64encode(iv + encryptor.tag + encrypted)
-        return encrypted
+        return encrypted.decode('utf-8')
 
     def decrypt(self, value):
         if isinstance(value, six.text_type):
@@ -208,12 +208,12 @@ class FernetEngine(EncryptionDecryptionBaseEngine):
             value = str(value)
         value = value.encode()
         encrypted = self.fernet.encrypt(value)
-        return encrypted
+        return encrypted.decode('utf-8')
 
     def decrypt(self, value):
         if isinstance(value, six.text_type):
             value = str(value)
-        decrypted = self.fernet.decrypt(value)
+        decrypted = self.fernet.decrypt(value.encode())
         if not isinstance(decrypted, six.string_types):
             decrypted = decrypted.decode('utf-8')
         return decrypted
@@ -344,7 +344,7 @@ class EncryptedType(TypeDecorator, ScalarCoercible):
 
     """
 
-    impl = LargeBinary
+    impl = String
 
     def __init__(self, type_in=None, key=None,
                  engine=None, padding=None, **kwargs):
--- a/sqlalchemy_utils/types/encrypted/encrypted_type.py
+++ b/sqlalchemy_utils/types/encrypted/encrypted_type.py
@@ -4,7 +4,7 @@ import datetime
 import os
 
 import six
-from sqlalchemy.types import LargeBinary, String, TypeDecorator
+from sqlalchemy.types import String, TypeDecorator
 
 from sqlalchemy_utils.exceptions import ImproperlyConfigured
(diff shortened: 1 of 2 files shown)


leosussan:
This seems to have broken things for me & probably will for others.

In my implementation (PostgreSQL 11, using `FernetEngine`), up to this point, encrypted values are stored as a `bytea` column, which is serialized into a `bytes` in Python, which of course has no `.encode()` and subsequently breaks [here](https://github.com/kvesteri/sqlalchemy-utils/pull/426/files#diff-59ab915885f53afdaa4ddbd7e0d15d4aR216).

This would conceivably break on any non-string object when extracted from the database.

Moving `.encode()` to the `str()` step & removing from value fixes extraction.

```python
        if isinstance(value, six.text_type):
            value = str(value).encode()
        decrypted = self.fernet.decrypt(value)
```

This doesn't solve the problem of encryption into the DB, which expects `bytea`, not a string.

aicioara:
I am sorry. I made this PR as a suggestion, I was not expecting it to be merged as-is, but I assumed things are alright for all use cases given the merge. It's been a while since March 14 when I opened it first, so I cannot remember much, but is there a way of making this work for both myself and @leosussan ?

leosussan:
I personally believe that your change is a good one, for the reasons you've outlined when you started this PR.

But - at the very least, the documentation / changelog should make clear that the change will break existing implementation.

For the sake of continuity, here's my current implementation. I had to subclass `EncryptedType`.

```python
class CustomEncryptedType(EncryptedType):
    impl = LargeBinary

    def process_bind_param(self, value, dialect):
        value = super().process_bind_param(value=value, dialect=dialect)
        if isinstance(value, str):
            value: bytes = value.encode()
        return value

    def process_result_value(self, value, dialect):
        if isinstance(value, bytes):
            value: str = value.decode()
        value: Optional[dict] = super().process_result_value(
            value=value, dialect=dialect
        )
        return value
```

dsimidzija:
Hey folks, this warning is not good on its own; it implies that we can easily switch from one type to another, but the switch is not backwards compatible, and this is not made obvious. A better message would be to "switch to ByteEncryptedType/LegacyEncryptedType to keep working with LargeBinary" in the future, or alternatively make a roadmap on how to actually migrate the db from EncryptedType to StringEncryptedType without losing any data in the process.

PR #429, opened 2020-03-27
Add test to show how to use type_cls and keyword args

pgiraud:
It replaces #255 (which I couldn't get working).
It doesn't add any code.

It just shows how CompositeType::type_cls is useful.
```
myType = CompositeType(
    'point',
    [
        sa.Column('x', sa.Integer),
        sa.Column('y', sa.Integer)
    ]
)
myType.type_cls(x=3, y=5)

point = dict(x=3, y=5)
myType.type_cls(**point)
```

This is very useful when you have a lot of columns and you want to create a new record from a dict and don't want to care about the column order.
Should it be documented?


pgiraud:
The failing test doesn't seem to have anything to do with my commit.

Issue #444, opened 2020-05-07
New enforcement of `String` over `LargeBinary` on EncryptedType breaks existing implementations


leosussan:
See my comment on https://github.com/kvesteri/sqlalchemy-utils/pull/426, which is when this was introduced.

leosussan:
Continuing the conversation from https://github.com/kvesteri/sqlalchemy-utils/pull/426#issuecomment-626853516:

> I personally believe that your change is a good one, for the reasons you've outlined when you started this PR.
> 
> But - at the very least, the documentation / changelog should make clear that the change will break existing implementation.
> 
> For the sake of continuity, here's my current implementation. I had to subclass `EncryptedType`.
> 
> ```python
> class CustomEncryptedType(EncryptedType):
>     impl = LargeBinary
> 
>     def process_bind_param(self, value, dialect):
>         value = super().process_bind_param(value=value, dialect=dialect)
>         if isinstance(value, str):
>             value: bytes = value.encode()
>         return value
> 
>     def process_result_value(self, value, dialect):
>         if isinstance(value, bytes):
>             value: str = value.decode()
>         value: Optional[dict] = super().process_result_value(
>             value=value, dialect=dialect
>         )
>         return value
> ```

villebro:
The change introduced by #426 breaks alembic migration code that works on MySQL on `sqlalchemy-utils==0.36.5`:

```python3
import sqlalchemy as sa
from alembic import op
from sqlalchemy_utils import EncryptedType

...

def upgrade():
    op.add_column(
        "dbs", sa.Column("password", EncryptedType(sa.String(1024)), nullable=True)
    )
```

```
  File "/opt/hostedtoolcache/Python/3.6.10/x64/lib/python3.6/site-packages/sqlalchemy/dialects/mysql/base.py", line 2047, in visit_VARCHAR
    "VARCHAR requires a length on dialect %s" % self.dialect.name
sqlalchemy.exc.CompileError: VARCHAR requires a length on dialect mysql
```

instantlinux:
Ditto this for me: this is a *breaking change* that would require a semver update and doc on what to do instead.  I get the following stacktrace on a migration that works fine with 0.36.4:
```
Traceback (most recent call last):
  File "/home/richb/apicrud/python_env/lib/python3.6/site-packages/sqlalchemy/sql/compiler.py", line 2897, in visit_create_table
    create_column, first_pk=column.primary_key and not first_pk
[...]

PR #450, opened 2020-05-23
Revert #426 and add support for Legacy Encrypted Type

rushilsrivastava:
This PR attempts to solve the problems outlined in #444 which were introduced in #426. The idea is to keep the old LargeBinary implementation and add a deprecated warning to it, as well as add support for the new `String` implementation for `EncryptedType`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,15 @@ Session.vim
 
 # Sublime Text
 *.sublime-*
+
+# JetBrains
+.idea/
+
+# Environments
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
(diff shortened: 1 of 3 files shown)


rushilsrivastava:
Currently using @leosussan's code from [this comment](https://github.com/kvesteri/sqlalchemy-utils/pull/426#issuecomment-626853516).

rushilsrivastava:
`EncryptedType` now works for my local migrations. Should throw a warning when using it as well? Can't confirm this works.

rushilsrivastava:
CC @kvesteri, it looks like the linter is having problems with code that hasn't been modified.

PR #463, opened 2020-07-09
fix database existence check

bernt-matthias:
follow up to https://github.com/kvesteri/sqlalchemy-utils/pull/372

TODO:
- [ ] always dispose engines (also for other dialects)
- [ ] just use the else branch for postgres?
- [ ] docs

### DIFF
--- a/sqlalchemy_utils/functions/database.py
+++ b/sqlalchemy_utils/functions/database.py
@@ -420,10 +420,12 @@ def is_auto_assigned_date_column(column):
     )
 
 
-def database_exists(url):
+def database_exists(url, postgres_db=['postgres', 'template0', 'template1',
+                                      None]):
     """Check if a database exists.
 
     :param url: A SQLAlchemy engine URL.
+    :postgres_db: TODO.
 
     Performs backend-specific testing to quickly determine if a database
     exists on the server. ::
@@ -461,9 +463,17 @@ def database_exists(url):
     database, url.database = url.database, None
     engine = sa.create_engine(url)
 
-    if engine.dialect.name == 'postgresql':
-        text = "SELECT 1 FROM pg_database WHERE datname='%s'" % database
-        return bool(get_scalar_result(engine, text))
+    if engine.dialect.name.startswith('postgresql'):
+        engine.dispose()
+        for pdb in postgres_db:
+            url.database = pdb
+            engine = sa.create_engine(url)
+            text = "SELECT 1 FROM pg_database WHERE datname='%s'" % database
+            try:
+                return bool(get_scalar_result(engine, text))
+            except OperationalError:
+                continue
+        return False
 
     elif engine.dialect.name == 'mysql':
         text = ("SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA "


CaselIT:
Other than the than the disposal of the engines, I think this should work fine 👍

sqlalchemy_utils/functions/database.py
```diff
@@ -459,41 +462,49 @@ def sqlite_file_exists(database):
 
     url = copy(make_url(url))
     database, url.database = url.database, None
-    engine = sa.create_engine(url)
+    engine = None
 
-    if engine.dialect.name == 'postgresql':
-        text = "SELECT 1 FROM pg_database WHERE datname='%s'" % database
-        return bool(get_scalar_result(engine, text))
+    if engine.dialect.name.startswith('postgresql'):
```
nsoranzo, line 467: Shouldn't this raise an `AttributeError` since `engine` is set to `None`?

bernt-matthias: true

bernt-matthias: would it be an alternatve to use `url.drivername` instead. Otherwise I would revert to creating the engine before the if-else.

nsoranzo: I wonder why the tests didn't fail?

I think you can set `dialect_name = url.get_dialect().name` before the `if` and then
```suggestion
    if dialect_name == 'postgresql':
```
and similarly for the following `elif`

bernt-matthias:
Wondering if one could extend a test case such that a postgres DB without CONNECT privilege for the postgres database is tested.

https://github.com/kvesteri/sqlalchemy-utils/blob/3090944c19aca76f57804637fee5788e8f530ece/tests/functions/test_database.py#L63

sqlalchemy_utils/functions/database.py
```diff
@@ -459,41 +462,50 @@ def sqlite_file_exists(database):
...
-        text = "SELECT 1 FROM pg_database WHERE datname='%s'" % database
-        return bool(get_scalar_result(engine, text))
-
-    elif engine.dialect.name == 'mysql':
+    if dialect_name == 'postgresql':
+        ret = False
+        for pdb in postgres_db:
+            url.database = pdb
+            engine = sa.create_engine(url)
+            text = "SELECT 1 FROM pg_database WHERE datname='%s'" % database
```
nsoranzo, line 473:
```suggestion
        text = "SELECT 1 FROM pg_database WHERE datname='%s'" % database
        for pdb in postgres_db:
            url.database = pdb
            engine = sa.create_engine(url)
```

sqlalchemy_utils/functions/database.py
```diff
@@ -420,10 +420,13 @@ def is_auto_assigned_date_column(column):
     )
 
 
-def database_exists(url):
+def database_exists(url, postgres_db=['postgres', 'template0', 'template1',
```
jtbeach, line 423: Note: default params should never be mutable objects in Python. It's a common gotcha: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

sqlalchemy_utils/functions/database.py
```diff
@@ -459,41 +462,50 @@ def sqlite_file_exists(database):
...
         text = "SELECT 1 FROM pg_database WHERE datname='%s'" % database
-        return bool(get_scalar_result(engine, text))
+        for pdb in postgres_db:
+            url.database = pdb
+            engine = sa.create_engine(url)
+            try:
+                ret = bool(get_scalar_result(engine, text))
+            except (ProgrammingError, OperationalError):
+                pass
+            engine.dispose()
```
jtbeach, line 478: Note this will always return the result from the last element in the `postgres_db` array. I think you probably want to break out pof this loop if `ret` is `True`

sqlalchemy_utils/functions/database.py
bernt-matthias: So it would be better to make it a tuple? Or `None`?

sqlalchemy_utils/functions/database.py
bernt-matthias: Of course, my bad. Thanks.

sqlalchemy_utils/functions/database.py
jtbeach: Best to default to None and then set to the current default list inside the function if the param is None

sqlalchemy_utils/functions/database.py
```diff
@@ -420,10 +420,13 @@ def is_auto_assigned_date_column(column):
     )
 
 
-def database_exists(url):
+def database_exists(url, postgres_db=('postgres', 'template0', 'template1',
+                                      None)):
```
nsoranzo, line 424:
```suggestion
def database_exists(url, postgres_db=None):
```

sqlalchemy_utils/functions/database.py
```diff
@@ -459,41 +462,52 @@ def sqlite_file_exists(database):
...
+    dialect_name = url.get_dialect().name
 
-    if engine.dialect.name == 'postgresql':
+    if dialect_name == 'postgresql':
+        ret = False
         text = "SELECT 1 FROM pg_database WHERE datname='%s'" % database
-        return bool(get_scalar_result(engine, text))
-
-    elif engine.dialect.name == 'mysql':
+        for pdb in postgres_db:
```
nsoranzo, line 471:
```suggestion
        if postgres_db is None:
            postgres_db = ('postgres', 'template0', 'template1')
        for pdb in postgres_db:
```

bernt-matthias: Should we keep `None`? This should use the default, i.e. connect to a db `USERNAME`?

nsoranzo: Sure, sorry, suggestion updated.

sqlalchemy_utils/functions/database.py
```diff
@@ -459,41 +461,54 @@ def sqlite_file_exists(database):
...
+            url.database = pdb
+            engine = sa.create_engine(url)
+            try:
+                ret = bool(get_scalar_result(engine, text))
+            except (ProgrammingError, OperationalError):
+                pass
+            engine.dispose()
+            if ret:
+                return ret
```
nsoranzo, line 481: I think we should exit the for loop also when `ret` gets set to `False`.
```suggestion
                ret = bool(get_scalar_result(engine, text))
                break
            except (ProgrammingError, OperationalError):
                pass
            finally:
                engine.dispose()
```

sqlalchemy_utils/functions/database.py
```diff
@@ -459,41 +461,54 @@ def sqlite_file_exists(database):
...
+        for pdb in postgres_db:
+            url.database = pdb
+            engine = sa.create_engine(url)
+            try:
+                ret = bool(get_scalar_result(engine, text))
+                break
+            except (ProgrammingError, OperationalError):
+                pass
+            finally:
+                engine.dispose()
```
jtbeach, line 481: Do we need to set engine to None here -- is it safe to call dispose() twice on an engine? A second dispose() will be called on line 510

bernt-matthias: true. how about this https://github.com/kvesteri/sqlalchemy-utils/pull/463/commits/e4815822ba7497afbf38891e8e1ca83b85e9f928

jtbeach: I think now we aren't disposing engines between each iteration of the loop if there are not exceptions. I'm not the author of this code, so don't want to offer too many opinions, but this code looks a little brittle in terms of the manual calls to Engine.dispose(). I would just use a context manager here...one that called dispose() in __exit__ and then use `with` everywhere I created an Engine

bernt-matthias: If I understand https://docs.sqlalchemy.org/en/14/core/connections.html?highlight=dispose#connectionless-execution-implicit-execution correctly, the current code uses implicit connections (i.e. calling execute on the engine). This may be changed to 

```
with engine.connect() as connection:
    connection.execute(...)
```

then we might no need to use `dispose` at all https://docs.sqlalchemy.org/en/14/core/connections.html?highlight=dispose#engine-disposal .. 

but I'm not an sqlalchemy expert at all .. but I'm fine with implementing the change if you agree

bernt-matthias: also dispose is called in `get_scalar_result` .. i.e. in the case of a successful existence check dispose is currently called twice... but according to the docs it just closes connections.

jtbeach: I would do this:

```

Comment (bernt-matthias) [reply sqlalchemy_utils/functions/database.py]: Thanks for the detailed explanations. I tried a slightly different way in https://github.com/kvesteri/sqlalchemy-utils/pull/463/commits/6b5cdb252c55d138e1335d6b64fcc3a1381681a8 

1. use a connection for the execution and close the connection (via `with`). Already with this there should be no need to dispose the engine (since there should be no open connections in the pool that need to be closed).
2. explicitly use the Null connection pool .. just to be sure.

Lets see if this passes tests.

**Comment by bernt-matthias** [comment]
Tests are passing. Will test now the postgres part with Galaxy.

**Comment by bernt-matthias** [comment]
Works.

File: sqlalchemy_utils/functions/database.py [hunk 9]
```diff
@@ -420,10 +421,12 @@ def is_auto_assigned_date_column(column):
     )
 
 
-def database_exists(url):
+def database_exists(url, postgres_db=None):
     """Check if a database exists.
 
     :param url: A SQLAlchemy engine URL.
+    :postgres_db: Only applies to postgres. List of databases to try to connect
```
kvesteri, line 428: Do this parameter apply to any other databases? In other words does for example Oracle or MSSQL have similar things that a default database should be provided in order to check the existence of another database. If so, we should use some different naming for this.

Furthermore the description of the parameter suggests this parameter is a list of values but the name does not. Either the parameter should be named ``postgres_databases`` (or something similar) OR this parameter could be string. 

I'm leaning towards the latter option as the end user who is using this function should know which default database to use and just provide that as a string parameter for this function.

bernt-matthias: > Do this parameter apply to any other databases?

I have no idea. But it seems to be postgres specific (I also found no issue that indicates such a problem for other databases). Still - if desired - I could change the parameter name  to `databases` and indicate in the docs that it currently only applies to postgres.

> Furthermore the description of the parameter suggests this parameter is a list of values but the name does not.

I would go for `postgres_databases`. The list has the advantage will work for many settings automatically.

sqlalchemy_utils/functions/database.py
```diff
@@ -442,11 +445,8 @@ def database_exists(url):
     """
 
     def get_scalar_result(engine, sql):
-        result_proxy = engine.execute(sql)
-        result = result_proxy.scalar()
-        result_proxy.close()
-        engine.dispose()
-        return result
+        with engine.connect() as conn:
+            return conn.scalar(sql)
```
ziima, line 449: Since it's only a two-liner, is the inline function still needed?

bernt-matthias: I would say so. Still avoids some code duplication. Do you have some alternative in mind?

ziima: That's true, but personally I don't really like dynamically defined functions. Maybe move it to the module level?

bernt-matthias: > personally I don't really like dynamically defined functions

Can do. Out of interest: is there a downside, apart from violating style preferences?

ziima: The function is created every time and that most likely eats up a bit of performance.

kvesteri:
The more I think about this the more I'm convinced that we should change the function signature of ``create_database(url)``, ``drop_database(url)`` and ``database_exists(url)`` to following:

``drop_database(engine_or_connection, database_name)``
``create_database(engine_or_connection, database_name)``
``drop_database(engine_or_connection, database_name)``

This would solve the issue raised in this PR. It would also solve #467 and it would have the benefit of being able to use existing connection / engine. What do you guys think?

bernt-matthias:
Thanks for the feedback @kvesteri . 

I guess I agree .. given also the code duplication for the engine creation in the current implementation of the three functions.

But I think one would also like to add a function to establish a connection (or engine) -- that could be passed to the three functions. Otherwise one would just delegate the "complexity" to establish a connection which is currently hidden in the three functions to the calling code.

Alternatively one could also add such a function and  call it in each of the three function -- just to remove the code duplication. This would have the advantage that the API is unchanged.

ziima:
I would definitely keep the current API, since the most common use case, in my opinion, is most likely to check/create/drop a database using the same connection string as is used in `create_engine`. I find that extremely helpful.

On the other hand, I agree that a new set of functions `*_database_internal(engine_or_connection, database_name)` could help to solve some of the issues and reduce some of the complexity of the current implementation.

kvesteri:
@ziima I'd like to hear more of your reasoning since I don't find it cumbersome at all to first create an engine / connection object and then re-use that in subsequent ``check / create / drop`` calls.

I do find the proposed solution quite cumbersome though with a postgresql specific keyword argument. The solution proposed in this PR does not solve #467 . It can't reuse an existing connection (= creates unnecessary connections). Furthermore I want the function signature to avoid database specific parameters unless absolutely necessary.

ziima:
In my use case I create a test database when running unittests, and as such I try to keep it as simple as possible.

Minimal working example
```python
import unittest

from sqlalchemy_utils.functions import (create_database, database_exists, drop_database)

TEST_DATABASE = 'postgresql:///test_db'

class MyTest(unittest.TestCase):
    def setUp(self):
        if not database_exists(TEST_DATABASE):
            create_database(TEST_DATABASE)

    def tearDown(self):
        drop_database(TEST_DATABASE)

    def test_foo(self):
        engine = create_engine(TEST_DATABASE)
        self.assertTrue(False)
```

I don't know much about other databases than sqlite3 and postgres, but using your proposal, the example would expanded significantly:
```python
import unittest
from copy import copy

from sqlalchemy.engine.url import make_url
from sqlalchemy_utils.functions import (create_database, database_exists, drop_database)

TEST_DATABASE = 'postgresql:///test_db'

class MyTest(unittest.TestCase):
    def setUp(self):
        test_database_url = make_url(TEST_DATABASE)
        self.test_db_name = test_database_url.database
        # Here the whole backend specific code would have been
        admin_database = copy(test_database_url)
        if test_database_url.drivername == 'postgresql':
            admin_database.database = 'template1'  # Still simple version, I ignore other possible database names and a case for #467
        # elif...
        self.admin_connection = create_engine(admin_database)
        if not database_exists(self.admin_connection, self.test_db_name):
            create_database(self.admin_connection, self.test_db_name)

    def tearDown(self):
        drop_database(self.admin_connection, self.test_db_name)

    def test_foo(self):
        engine = create_engine(TEST_DATABASE)
        self.assertTrue(False)
```
As @bernt-matthias noted above, the complexity would just have to be handled before the functions are called. That's not very handy.

Regarding the problems you mentioned, they might be easier to manage if we create some sort of `DatabaseManager` with methods `check`, `create` and `drop`. It could maintain a connection between the subsequent operations. The current API could be kept for simple cases, where you don't care about connection reusability.

kvesteri:
@ziima the problems you referred to in your comment seem to be limitations of UnitTest. With pytest for example, one could just use fixtures and reuse connection fixture on subsequent function calls. Thus I'd like to see this changed so that the functions would take connection / engine as the first parameter. Introducing DatabaseManager seems counter-intuitive and a bit clumsy.

bernt-matthias:
Thanks for coming back to this @kvesteri 

Sounds like a larger change and I'm not sure if I will be able to implement this since actually I have never used the sqlalchemy library as a programmer (it's just used in a project that I'm involved in). So I have for instance no idea of the concepts behind an engine. 

With a more precise plan I might give it a go. Maybe something like pseudo code for the at least one of the functions...

tests/types/test_encrypted.py
```diff
@@ -153,8 +153,8 @@ def user_datetime():
 
 @pytest.fixture
 def test_token():
-    import string
     import random
+    import string
```
nsoranzo, line 157: Shouldn't this standard library imports just go to the top of the file?

sqlalchemy_utils/functions/database.py
```diff
@@ -441,59 +459,46 @@ def database_exists(url):
...
+    dialect_name = url.get_dialect().name
 
-    if engine.dialect.name == 'postgresql':
+    if dialect_name == 'postgresql':
         text = "SELECT 1 FROM pg_database WHERE datname='%s'" % database
-        return bool(get_scalar_result(engine, text))
-
-    elif engine.dialect.name == 'mysql':
+        if databases is None:
+            databases = ('postgres', 'template0', 'template1', None)
```
bsquizz, line 469: If `url` passed into this function provided a db name, for example:
```
postgresql://user:pass@1.2.3.4/dbname
```

Wouldn't we want to check `dbname` first here?

bsquizz: Ok nevermind, I see that the SQL command provided by `text` will handle that.

nsoranzo:
FYI, in the https://github.com/nsoranzo/sqlalchemy-utils/tree/sqlalchemy14 branch I'm trying to combine this PR with #487 (and more).

PR #464, opened 2020-07-09
Fix #459 - Check inner type of the scalar list

### LINKED ISSUE #459: ScalarListType doesn't coerce input values
`ScalarListType` may store values which it cannot return. This occurs if values which can't be processed by a `coerce_func` are set as input.

An example with UUID:
```python
from uuid import UUID
from sqlalchemy_utils import ScalarListType

engine = create_engine('sqlite:////:memory:')

field = ScalarListType(UUID)

uuid = UUID(int=42)
serialized = field.process_bind_param([uuid], engine.dialect)
#>>> '00000000-0000-0000-0000-00000000002a'
assert [uuid] == field.process_result_value(serialized, engine.dialect)
#>>> [UUID('00000000-0000-0000-0000-00000000002a')]

serialized = field.process_bind_param([str(uuid)], engine.dialect)
#>>> '00000000-0000-0000-0000-00000000002a'
assert [uuid] == field.process_result_value(serialized, engine.dialect)
#>>> [UUID('00000000-0000-0000-0000-00000000002a')]

invalid_uuid = 'invalid-uuid'
inv_serialized = field.process_bind_param([invalid_uuid], engine.dialect)
#>>> 'invalid-uuid'
assert [invalid_uuid] == field.process_result_value(inv_serialized, engine.dialect)
#*** ValueError: badly formed hexadecimal UUID string
```

In my opinion the main issue is that `ScalarListType` doesn't coerce values on input to check it will be able to return them as well.

ziima:
A complex fix for #459. I've changed the first argument of `ScalarListType` to a type of its values, since in most cases it's the same as `coerce_func`. All values are checked if they match the type on save. Backwards compatibility is maintained.


sqlalchemy_utils/types/scalar_list.py
```diff
@@ -50,13 +51,31 @@ class Player(Base):
...
+        Separator of the values. Default is ``,``.
+    :param coerce_func:
+        Custom function to coerce values when read from database.
+        By default ``inner_type`` is used instead.
     """
 
     impl = sa.UnicodeText()
 
-    def __init__(self, coerce_func=six.text_type, separator=u','):
+    def __init__(self, inner_type=six.text_type, separator=u',',
```
kvesteri, line 65: Why not keeping the parameter order as it was and adding the new parameter as the last parameter? That way we wouldn't need to throw any warnings.

ziima: I was thinking about that. I believe, having `inner_type` defined is the only way I can ensure validation of the values before they're written to database. So I made `inner_type` primary and `coerce_func` optional. It would be weird to have an optional parameter first.

ziima: @kvesteri Any decision?

ziima: @kvesteri I'd like to move this forward. Can I keep the change, or should I add the `inner_type` as the last parameter?

PR #473, opened 2020-09-07
Add Py38 Tests

yangchoo:


yangchoo:
fwiw, not really sure what's going on with the lint setup for isort.

I ended up selectively fixing it as I could not replicate the CI results locally even with a tox runner with matching `isort` and python versions.

PR #479, opened 2020-10-22
fix: export StringEncryptedType

jd:
The new type is not as easy to import as the deprecated EncryptedType class.


kurtmckee:
@kvesteri This PR can be closed because the changes already exist in sqlalchemy-utils.

PR #481, opened 2020-11-13
Sqlalchemy 1.4 support

AlekseevAV:


kvesteri:
The tests are failing. Also the imports should be conditional in a sense that older SQLAlchemy versions are still supported. SA 1.4 should be added to the test matrix.

PR #483, opened 2020-11-29
Added support for SA 1.14. Applied _ColumnEntity import for SA 1.14. …

slipovenko:
…Added adapter for URL to database_exists, create_database, drop_database


Arengar:
an attempt to support SA 1.14
The main issue that now SA's URL object is not an object but the namedtuple instanse. So in order to support the wanted behaviour had to introduce the Namespace object in database_exists, create_database, drop_database funtions

PR #484, opened 2020-12-04
Fix #319

### LINKED ISSUE #319: CompositeType example doesn't work

StdCarrot:
- Fixed problem: 
typemap doesn't exist error when trying to use sub-column of CompositeType in select query(equivalent as get Comparator of sub-column)

- Solving by:
Get sub-column's type correctly.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,12 @@ Changelog
 Here you can see the full list of changes between each SQLAlchemy-Utils release.
 
 
+0.38.3 (2022-xx-xx)
+^^^^^^^^^^^^^^^^^^^
+
+- Fixed double-quoted UUID's in sqlalchemy >= 1.4.30 (#581, pull request courtesy of kurtmckee)
+
+
 0.38.2 (2021-12-29)
 ^^^^^^^^^^^^^^^^^^^
 
--- a/sqlalchemy_utils/__init__.py
+++ b/sqlalchemy_utils/__init__.py
@@ -101,4 +101,4 @@ from .view import (  # noqa
     refresh_materialized_view
 )
 
-__version__ = '0.38.2'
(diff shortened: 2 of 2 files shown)


kurtmckee:
@StdCarrot Thanks for your work on this! I'm triaging to help with the sqlalchemy-utils project and see that this PR needs a unit test to demonstrate what's broken and how it's fixed.

Are you able to return to this PR? If so, great! Please respond back if you need help creating a unit test. 👍

PR #487, opened 2020-12-26
Add support for SQLAlchemy 1.4

cnu:
Import _ColumnEntity from sqlalchemy.orm.context if importing from
.query fails.

And while checking if database_exists, use url.set() as URL object is
now immutable.

### DIFF
--- a/sqlalchemy_utils/functions/database.py
+++ b/sqlalchemy_utils/functions/database.py
@@ -458,7 +458,11 @@ def database_exists(url):
         return header[:16] == b'SQLite format 3\x00'
 
     url = copy(make_url(url))
-    database, url.database = url.database, None
+    try:
+        database, url.database = url.database, None
+    except AttributeError:  #SQLalchemy 1.4: url is immutable
+        database = url.database
+        url.set(database=None)
     engine = sa.create_engine(url)
 
     if engine.dialect.name == 'postgresql':
--- a/sqlalchemy_utils/functions/orm.py
+++ b/sqlalchemy_utils/functions/orm.py
@@ -11,7 +11,10 @@ from sqlalchemy.orm import mapperlib
 from sqlalchemy.orm.attributes import InstrumentedAttribute
 from sqlalchemy.orm.exc import UnmappedInstanceError
 from sqlalchemy.orm.properties import ColumnProperty, RelationshipProperty
-from sqlalchemy.orm.query import _ColumnEntity
+try:
+    from sqlalchemy.orm.query import _ColumnEntity
+except ImportError:  # SQLAlchemy 1.4
+    from sqlalchemy.orm.context import _ColumnEntity
 from sqlalchemy.orm.session import object_session
 from sqlalchemy.orm.util import AliasedInsp
 


jpic:
Nice! Works here :tophat:

jpic:
I had to change this:

```
--- a/sqlalchemy_utils/functions/database.py
+++ b/sqlalchemy_utils/functions/database.py
@@ -462,7 +462,7 @@ def database_exists(url):
     except AttributeError:  #SQLalchemy 1.4: url is immutable
         database = url.database
-        url.set(database=None)
+        url.set(database='')
     engine = sa.create_engine(url)
 
     if engine.dialect.name == 'postgresql':
```

Because:

```
(Pdb) url.set(database=None)
postgresql://localhost/example_project
(Pdb) url
postgresql://localhost/example_project
(Pdb) url.set(database=False)
*** TypeError: database must be a string
(Pdb) url.set(database='')
postgresql://localhost/
```

jpic:
Actually, I had to make more changes than I thought:

```
diff --git a/sqlalchemy_utils/functions/database.py b/sqlalchemy_utils/functions/database.py
index c80620d..6fd6f6a 100644
--- a/sqlalchemy_utils/functions/database.py
+++ b/sqlalchemy_utils/functions/database.py
@@ -462,7 +462,10 @@ def database_exists(url):
     except AttributeError:  #SQLalchemy 1.4: url is immutable
         database = url.database
-        url.set(database=None)
+        if url.drivername.startswith('postgres'):
+            url = url.set(database='postgres')
+        else:
+            url = url.set(database=None)
     engine = sa.create_engine(url)

     if engine.dialect.name == 'postgresql':
@@ -527,7 +530,10 @@ def create_database(url, encoding='utf8', template=None):
     database = url.database

     if url.drivername.startswith('postgres'):
-        url.database = 'postgres'
+        try:
+            url.database = 'postgres'
+        except AttributeError:
+            url = url.set(database='postgres')
     elif url.drivername.startswith('mssql'):
         url.database = 'master'
     elif not url.drivername.startswith('sqlite'):
```

I suppose more should be done to support other database drivers ?

jpic:
Also:

```diff
commit be1ad6bcf979ccb9f43c35ef82e883deaca5641b (HEAD -> sqlalchemy14, origin/sqlalchemy14)
Author: jpic <jpic@yourlabs.org>
Date:   Sun Jan 17 00:02:23 2021 +0100

    Columns does not return a dict in SQLA1.4

diff --git a/sqlalchemy_utils/functions/orm.py b/sqlalchemy_utils/functions/orm.py
index daa2bf4..e1953b8 100644
--- a/sqlalchemy_utils/functions/orm.py
+++ b/sqlalchemy_utils/functions/orm.py
@@ -370,7 +370,7 @@ def get_primary_keys(mixed):
     """
     return OrderedDict(
         (
-            (key, column) for key, column in get_columns(mixed).items()
+            (key, column) for key, column in dict(get_columns(mixed)).items()
             if column.primary_key
         )
     )
diff --git a/sqlalchemy_utils/listeners.py b/sqlalchemy_utils/listeners.py
index 8213272..0a2e684 100644
--- a/sqlalchemy_utils/listeners.py
+++ b/sqlalchemy_utils/listeners.py
@@ -22,7 +22,7 @@ def coercion_listener(mapper, class_):


 def instant_defaults_listener(target, args, kwargs):
-    for key, column in sa.inspect(target.__class__).columns.items():
+    for key, column in dict(sa.inspect(target.__class__).columns).items():
         if hasattr(column, 'default') and column.default is not None:
             if callable(column.default.arg):
                 setattr(target, key, column.default.arg(target))
```

sqlalchemy_utils/functions/database.py
```diff
@@ -458,7 +458,11 @@ def sqlite_file_exists(database):
         return header[:16] == b'SQLite format 3\x00'
 
     url = copy(make_url(url))
-    database, url.database = url.database, None
+    try:
+        database, url.database = url.database, None
+    except AttributeError:  #SQLalchemy 1.4: url is immutable
+        database = url.database
+        url.set(database=None)
```
ic4f, line 465: Unfortunately, this won't work because `url.set()` [returns a *new* URL object](https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.engine.URL.set): it does not modify `url`.
However, `url = url.set(database=None)` won't work either because `None` cannot be assigned in the [method's current implementation](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/engine/url.py#L262).

jssuzanne: Hi,

For my test I added a wrapper to limit the impact in the code

Example in database_exists
```
    class UrlWrapper:
        url = {}

        def __init__(self, url):
            self.__class__.url[id(self)] = url

        def __getattr__(self, key, **kw):
            return getattr(self.url[id(self)], key, **kw)

        def __setattr__(self, key, value):
            try:
                setattr(self.url[id(self)], key, value)
            except AttributeError:
                if value is None:
                    value = ''

                self.__class__.url[id(self)] = self.url[id(self)].set(
                    **{key: value})

        def __call__(self):
            return self.url[id(self)]

    url = UrlWrapper(copy(make_url(url)))

    engine = sa.create_engine(url())

```

ic4f: Yes, this will modify the object's attribute. But you may not really *need* to modify it. There's an explanation of the decision to make the URL object immutable in [SQLAlchemy's documentation](https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#the-url-object-is-now-immutable). Setting it to `''` works, but the semantics feel off: we don't want to set the database to an empty string, we want to *unset* it.

sqlalchemy_utils/functions/orm.py
```diff
@@ -11,7 +11,10 @@
 from sqlalchemy.orm.attributes import InstrumentedAttribute
 from sqlalchemy.orm.exc import UnmappedInstanceError
 from sqlalchemy.orm.properties import ColumnProperty, RelationshipProperty
-from sqlalchemy.orm.query import _ColumnEntity
+try:
+    from sqlalchemy.orm.query import _ColumnEntity
+except ImportError:  # SQLAlchemy 1.4
+    from sqlalchemy.orm.context import _ColumnEntity
```
nsoranzo, line 17: Swapping the 2 imports would be more future-proof.

lucventurini:
The attached patch fixes the broken unit-test, by fixing the remaining url-setting functions in `sqlalchemy_utils/functions/database.py`.


[sqla1.4.patch.txt](https://github.com/kvesteri/sqlalchemy-utils/files/6151147/sqla1.4.patch.txt)

sqlalchemy_utils/functions/database.py
lucventurini: @ic4f

I think I have found a solution to that, please see here:

https://github.com/lucventurini/sqlalchemy-utils/commit/1f095582d0449700367a4d20e9e265074d023116

In short, I wrote a wrapper function that creates a new "URL" object (in SA>=1.4) with the database appropriately set.

Kind regards,

nsoranzo:
FYI, in the https://github.com/nsoranzo/sqlalchemy-utils/tree/sqlalchemy14 branch I'm trying to combine this PR and @lucventurini's follow-up commits with https://github.com/kvesteri/sqlalchemy-utils/pull/463 (and more).

PR #503, opened 2021-03-06
Handle missing pymysql import

ic4f:
This applies the same fix (no modification) already present in `tests/functions/database.py` ([here](https://github.com/kvesteri/sqlalchemy-utils/blob/master/tests/functions/test_database.py#L8) and [here](https://github.com/kvesteri/sqlalchemy-utils/blob/master/tests/functions/test_database.py#L44)).

This fixes local test runs on systems where pymysql, as well as mysql, are not installed.

Summary of edits:
1. Wrap import of pymysql package in try/except; 
2. Skip test if pymysql is missing.

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -7,6 +7,12 @@ from sqlalchemy_utils import (
     refresh_materialized_view
 )
 
+pymysql = None
+try:
+    import pymysql  # noqa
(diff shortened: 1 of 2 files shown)


ic4f:
I was mistaken, the dependency is listed in setup.py. Closing.

PR #508, opened 2021-04-08
selectable.c -> selectable.selected_columns

doriantaylor:
Hi, I was getting deprecation errors in my tests with `sa` 1.4.5; figured I might as well share this little patch with you.

Note I couldn't run your test suite because of that call to a missing `sqlalchemy.orm.query._ColumnEntity`, and I haven't the foggiest how to fix that so I couldn't test this.


nsoranzo:
@doriantaylor An alternative patch has been proposed in commit [`f0020cf` (#506)](https://github.com/kvesteri/sqlalchemy-utils/pull/506/commits/f0020cf8ad7295a9d4edd36f0533c105f53e6823) .

doriantaylor:
Ah cool; when do you plan to ship?

PR #509, opened 2021-04-12
Adding default_engine Class var

craig-rueda:
Adds new class variable `default_engine` to `StringEncryptedType` to allow for overriding of default behavior. Currently, `StringEncryptedType`'s ctor is hard-coded to utilize the `AesEngine` if no other engine is specified. This PR allows the default engine to be configured during app bootstrap.


kurtmckee:
@craig-rueda It seems like it's already possible to override default behavior by specifying the encryption engine when instantiating the class.

@kvesteri I recommend closing this PR. Would you review and make a determination?

kvesteri:
I see the reason for this kind of functionality. However there are merge conflicts and this PR is missing unit tests.

PR #510, opened 2021-04-16
Add 'zoneinfo' backend to TimezoneType

huonw:
This adds support for using the standard library's 'zoneinfo' module
as a backend for the TimezoneType. This module is available on Python
3.9+. For older versions, the backports.zoneinfo
module (https://github.com/pganssle/zoneinfo) is automatically used,
and is thus required on these versions.

(Thank you for sqlalchemy-utils, it's simplified a bunch of stuff very well, for us!)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,6 +35,7 @@ extras_require = {
         'pg8000>=1.12.4',
         'pytz>=2014.2',
         'python-dateutil>=2.6',
+        'backports.zoneinfo;python_version<"3.9"',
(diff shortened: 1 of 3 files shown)


huonw:
Hi @kvesteri, if you have a chance, we'd love to see this progress. Could you take a look? Thanks again for sqlalchemy-utils!

kvesteri:
Thanks for the PR! 🚂

Issue #522, opened 2021-05-10
Python 2.7 support broken


sydneypemberton1986:
https://sqlalchemy-utils.readthedocs.io/en/latest/installation.html#supported-platforms claims 2.7 support however there are MyPy type annotations in `database.py` that are only supported by Python 3 or greater. 

```
ipython 
Python 2.7.17 (default, Dec 12 2019, 14:22:11)
Type "copyright", "credits" or "license" for more information.

IPython 5.10.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import sqlalchemy_utils as l
  File "/Users/spemberton/AB_CORE/lib/python2.7/site-packages/sqlalchemy_utils/functions/database.py", line 423
    def _set_url_database(url: sa.engine.url.URL, database):
                             ^
SyntaxError: invalid syntax
```

Suggested fix: switch to Python-2 style comment-based type annotations.

sydneypemberton1986:
I'm on 0.37.0 btw.

After making the above fix there is an issue with `collections.abc`.

kvesteri:
Python 2.7 is no longer supported.

gustawdaniel:
there is solution for `collections.abc` problem https://github.com/dcantrell/pyparted/commit/447a7c5bc0b8038f2cf9865271acb08655c55464

which is last version with python2 support?

gustawdaniel:
I sound that `0.36.6` in issue https://github.com/kvesteri/sqlalchemy-utils/issues/461

PR #523, opened 2021-05-10
Fix for https://github.com/kvesteri/sqlalchemy-utils/issues/522

sydneypemberton1986:
I removed the type annotation and used a try/catch around `collections.abc`. 

I'm not familiar with your code process, but hopefully this gets you most of the way there.


kvesteri:
Python 2.7 is no longer supported, please see release notes.

PR #525, opened 2021-05-11
Add cache_ok=True class var to UUIDType

### LINKED ISSUE #524: TypeDecorator UUIDType() will not produce a cache key because the `cache_ok` flag is not set to True
When using SQLAlchemy 1.4.14 that supports [cache_ok](https://docs.sqlalchemy.org/en/14/core/custom_types.html#sqlalchemy.types.TypeDecorator.cache_ok), combined with `UUIDType` from SQLAlchemy-Utils, one will get the following `SAWarning`:
```
SAWarning: TypeDecorator UUIDType() will not produce a cache key because the ``cache_ok`` flag is not set to True. 
Set this flag to True if this type object's state is safe to use in a cache key, or False to disable this warning.
```

From what I can gather this cannot be fixed by an end-user as the `cache_ok` field should be added to `UUIDType()` class within SQLALchemy-Utils.

bibajz:
Resolves #524

### DIFF
--- a/sqlalchemy_utils/types/uuid.py
+++ b/sqlalchemy_utils/types/uuid.py
@@ -26,9 +26,11 @@ class UUIDType(types.TypeDecorator, ScalarCoercible):
     """
     impl = types.BINARY(16)
 
+    cache_ok = True
+
     python_type = uuid.UUID
 
-    def __init__(self, binary=True, native=True, **kwargs):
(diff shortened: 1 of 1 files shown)


kvesteri:
Fixed by https://github.com/kvesteri/sqlalchemy-utils/pull/528

PR #557, opened 2021-11-01
wip: Avoid sqlalchemy URL deprecation warning

simonschmidt:
> sqlalchemy.exc.SADeprecationWarning: Calling URL() directly is deprecated and will be disabled in a future release.  The public constructor for URL is now the URL.create() method.

Not sure what happens with sqlalchemy 1.3 - this PR is mostly to run the test suite

### DIFF
--- a/sqlalchemy_utils/functions/database.py
+++ b/sqlalchemy_utils/functions/database.py
@@ -480,7 +480,6 @@ def database_exists(url):
 
     """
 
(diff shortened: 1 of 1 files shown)


gmontanola:
I tested things locally and just removing `copy` avoids the warning.

A simple `make_url(url)`. Do you know why `copy` is used?

PR #559, opened 2021-11-18
Fix ignored test cases

SuibianP:
Tests under this class seem not to be run due to [pytest discovery rules](https://docs.pytest.org/en/6.2.x/goodpractices.html#conventions-for-python-test-discovery), which requires classes to be prefixed with `Test`.

This also seems to be hiding potential problems with non-PostgreSQL database systems.

### DIFF
--- a/tests/types/test_date_range.py
+++ b/tests/types/test_date_range.py
@@ -42,7 +42,7 @@ def init_models(Booking):
 
 
 @pytest.mark.skipif('intervals is None')
-class DateRangeTestCase:
+class TestDateRange:
     def test_nullify_range(self, create_booking):
         booking = create_booking(None)
         assert booking.during is None


kurtmckee:
@SuibianP, thanks for your work on this. You were right -- making the tests discoverable has indeed revealed some issues. For posterity, here's the output from one of the failing tests on Python 3.7:

```
AttributeError("'DateInterval' object has no attribute 'lower_inc'")
```

Is this something you're comfortable debugging and resolving as a part of this PR?

(By the way, the Python 3.6 deprecation failures are resolved in the most recent commits. If you rebase those should disappear and let you run the tests in Python 3.6.)

SuibianP:
This issue roots from https://github.com/kvesteri/intervals/pull/49, and will be automatically solved by bumping the dependency version once that PR is merged.

On a separate note, the practice of de/serialising with `__str__` and the constructor is presumably fragile by nature IMHO…

PR #562, opened 2021-12-05
Fix #561, allow merge_references with mapped properties

### LINKED ISSUE #561: get_foreign_key_values fail when object attributes are mapped to different names
I tried to use `merge_references`, but it will fail inside of `get_foreign_key_values` with an error like this:

> `old_values = get_foreign_key_values(fk, from_)`
> File "...venv/lib64/python3.9/site-packages/sqlalchemy_utils/functions/foreign_keys.py", line 15, in `get_foreign_key_values`
>  `return dict(`
> File "...venv/lib64/python3.9/site-packages/sqlalchemy_utils/functions/foreign_keys.py", line 18, in <genexpr>
>  `getattr(obj, element.column.key)`
> **AttributeError: 'User' object has no attribute 'id'**

My `User` and for example a referring class looks somewhat like this:

```python
class User(...):
    __tablename__ = "user"
    userid: str = db.Column(db.String(30), nullable=False)
    # ...
    id_ = db.Column("id", Serial, primary_key=True)

class _UserAttribute(...):
    __tablename__ = "user_attribute"
    id = db.Column("id", Serial, primary_key=True)
    user: User = db.Column("user", Serial, db.ForeignKey("user.id"), nullable=False)
    name: str = db.Column(db.String(30))
    value: any = db.Column(db.PickleType(protocol=4))
```

This error is related to https://github.com/kvesteri/sqlalchemy-utils/blob/a4154bd0809bc6bbf0c27d5f7c0f3f2872edd779/sqlalchemy_utils/functions/foreign_keys.py#L18 as the code tries to access `id` on the `User`, but of cause it can not find it as it is named `id_`.
I guess you would have to test for a mapper like

```python
def get_foreign_key_values(fk, obj):
  mapper = inspect(obj.__class__)
  # ...
    getattr(obj, mapper.get_property_by_column(element.column).key)
```

I am not sure if this works without a mapping assigned or if it works at all.

susnux:
This fixes #561 

I am not sure if `synchronize_session='fetch'` is a good thing to do, maybe `session.expire_all` is a better thing to do (or commit the session).

### DIFF
--- a/sqlalchemy_utils/functions/foreign_keys.py
+++ b/sqlalchemy_utils/functions/foreign_keys.py
@@ -12,15 +12,17 @@ from .orm import _get_class_registry, get_column_key, get_mapper, get_tables
 
 
 def get_foreign_key_values(fk, obj):
+    mapper = get_mapper(obj)
     return dict(
         (
-            fk.constraint.columns.values()[index].key,
+            fk.constraint.columns.values()[index],
             getattr(obj, element.column.key)
+            if hasattr(obj, element.column.key)
+            else getattr(
+                obj, mapper.get_property_by_column(element.column).key
+            ),
         )
-        for
-        index, element
-        in
-        enumerate(fk.constraint.elements)
+        for index, element in enumerate(fk.constraint.elements)
     )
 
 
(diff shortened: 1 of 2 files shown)


kurtmckee:
@susnux Thanks for your work on this!

Would you rebase on the master branch and force push the changes here? I'd like to run CI against this before reviewing and merging. Thanks!

susnux:
@kurtmckee rebased and force-pushed

kurtmckee (approved):
Looks good to me!

PR #565, opened 2021-12-13
Fix cache_ok on ScalarCoercible subclasses

ziima:
Almost all of the `ScalarCoercible` subclasses still produced the warning about `cache_ok`. The `cache_ok` have been set on the `ScalarCoercible` instead of each final type class, as described in SQLAlchemy docs, see https://docs.sqlalchemy.org/en/14/errors.html#assertion-attributes-for-caching.

### DIFF
--- a/sqlalchemy_utils/types/scalar_coercible.py
+++ b/sqlalchemy_utils/types/scalar_coercible.py
@@ -1,6 +1,4 @@
 class ScalarCoercible:
-    cache_ok = True
-
     def _coerce(self, value):
         raise NotImplementedError
 
--- a/tests/types/test_date_range.py
+++ b/tests/types/test_date_range.py
@@ -82,6 +82,12 @@ class DateRangeTestCase:
         assert booking.during.lower == datetime(2015, 1, 1).date()
         assert booking.during.upper == datetime(2015, 1, 1).date()
 
+    def test_compilation(self, session, Booking):
+        query = sa.select([Booking.during])
+
+        # the type should be cacheable and not throw exception
+        session.execute(query)
+
 
 @pytest.mark.usefixtures('postgresql_dsn')
 class TestDateRangeOnPostgres(DateRangeTestCase):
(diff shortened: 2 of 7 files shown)


thnee:
Thanks for posting this, this branch definitely fixes the problem for me.

I started getting this error after upgrading from SQLAlchemy 1.4.27 to 1.4.28 or 1.4.29.

```
SAWarning: TypeDecorator ChoiceType(length=255) will not produce a cache key because the ``cache_ok`` attribute is not set to True.  This can have significant performance implications including some performance degradations in comparison to prior SQLAlchemy versions.  Set this attribute to True if this type object's state is safe to use in a cache key, or False to disable this warning. (Background on this error at: https://sqlalche.me/e/14/cprf)
```

I don't really understand much of what is going on here, but what this branch is doing definitely helps.

kurtmckee:
@ziima, thanks for your work on this! Are you able to return to this PR and update it? There are several things that need to be addressed:

1. There are now merge conflicts with the master branch.
2. The unit tests in sqlalchemy-utils have been changed so warnings are escalated to errors. It's likely that your branch no longer needs to filter warnings, as they will be automatically escalated to errors.
3. The unit tests did not pass in CI back in December.

Unfortunately the CI logs have been lost to time, but I think they'll re-run automatically when you push changes to resolve the merge conflict.

ziima:
It took me a while to get back. The branch is rebased and fixed. The CI fails though, because a new version of `cryptography` showed up and that causes warnings. But that's not related to this issue.

kurtmckee:
@ziima Thanks for the update!

I checked out your branch and ran it with and without the `cache_ok = True` line in the base `ScalarCoercible` class but didn't observe a change in the test results. It may be that something needs to be added to the unit tests to trigger the warning, or something was modified in sqlalchemy-utils that resolved this issue already.

Would you double-check the unit tests and confirm whether they need to be expanded to trigger the warning, or if this bug has been fixed by other changes to the code?

ziima:
@kurtmckee The bug has indeed already been fixed by another PR. Now, only the tests remain in here. They fail, if `cache_ok` is removed from the actual field types.

The removal of `cache_ok` from `ScalarCoercible` is a bit of a leftover from the original change. Since the `ScalarCoercible` is not actually a final field type, the `cache_ok` flag is ignored and thus it is useless and can/should be deleted.

kurtmckee:
Ah, got it. I confirmed that removing `cache_ok = True` from one of the final types triggered failures. Thanks for your work on this!

PR #566, opened 2021-12-14
Fix for nested composite types and allow schema specification

na-ka-na:
Tested by creating a nested composite type in a custom schema


kvesteri:
Tests are missing, all PRs should have tests in order to get merged

PR #573, opened 2022-01-05
Avoided deprecation warning for URL in SQLAlchemy 1.4+.

ngnpope:
Closes #557 which didn't take the following into consideration:

- Still need `make_url()` to convert connection strings to `URL`.
- Still need `copy()` for SQLAlchemy < 1.4 where `URL` is mutable.

Note that SQLAlchemy has implemented `__copy__` and `__deepcopy__` so that `URL.create()` is used instead of `URL()`:

- https://github.com/sqlalchemy/sqlalchemy/issues/7400

Although this patch is only required for SQLAlchemy >=1.4,<1.4.28 it makes sense to avoid copying unnecessarily.

### DIFF
--- a/sqlalchemy_utils/functions/database.py
+++ b/sqlalchemy_utils/functions/database.py
@@ -438,6 +438,7 @@ def _set_url_database(url: sa.engine.url.URL, database):
             query=url.query
         )
     else:  # SQLAlchemy <1.4
+        url = copy(url)
         url.database = database
         ret = url
     assert ret.database == database, ret
@@ -480,7 +481,7 @@ def database_exists(url):
 
     """
 
-    url = copy(make_url(url))
+    url = make_url(url)
     database = url.database
     dialect_name = url.get_dialect().name
     engine = None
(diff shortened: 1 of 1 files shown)


kvesteri:
I couldn't reproduce the warning in tests. Can you add a test case that reproduces the warning?

ngnpope:
> I couldn't reproduce the warning in tests. Can you add a test case that reproduces the warning?

The problem is that this only affects SQLAlchemy >=1.4,<1.4.28. `URL.__copy__()` was added in 1.4.28 which fixed this, so it's unlikely we'll be able to write a test for the deprecation warning.

kurtmckee:
@kvesteri I'll take a crack at making this reproducible as a break/fix chain of commits, at least with a local tox run.

kurtmckee:
@kvesteri I've created PR #587 that introduces a tox environment to demonstrate the DeprecationWarnings that @ngnpope's commits fix. Please review when you have an opportunity. :grinning:

kurtmckee:
This PR can be closed. #587, which cherry-picks the commits from @ngnpope, has been merged. :+1:

PR #582, opened 2022-03-17
Prevent TypeError exceptions involving JSONType

### LINKED ISSUE #543: JSONType should specify hashable=false
various JSON things are not hashable. For instance lists and dicts. So the JSONType should let SQLAlchemy know that so it doesn't do things like `if value in some_dict` where value is a JSONType.

To fix this should have a `hashable = False` here:
https://github.com/kvesteri/sqlalchemy-utils/blob/b5b48b13a22aa6d796e98c0225f367a2f010c788/sqlalchemy_utils/types/json.py#L58

This causes certain use cases for JSONType to break in SQLAlchemy 1.4.19 and up. 

For background and a MCVE of the bug, see: https://github.com/sqlalchemy/sqlalchemy/issues/6924

kurtmckee:
Closes #543.

Thanks to @jessehersch for providing code to reproduce the error!

### DIFF
--- a/sqlalchemy_utils/types/json.py
+++ b/sqlalchemy_utils/types/json.py
@@ -49,6 +49,7 @@ class JSONType(sa.types.TypeDecorator):
         session.commit()
     """
     impl = sa.UnicodeText
+    hashable = False
     cache_ok = True
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


kurtmckee:
_Rebased on master; tidied the tox configuration and added comments_

kurtmckee:
_Rebased on master_

PR #584, opened 2022-03-18
Feature/pg composite mixed case support

### LINKED ISSUE #583: CompositeType does not allow mixed case name
In our production database for which we are trying to build a sqlalchemy interface, there are existing composite types defined that have mixed cases.

Currently, `CompositeType` class has this line:
https://github.com/kvesteri/sqlalchemy-utils/blob/850d0ea5a8a84c0529d175ee41b9036bea597119/sqlalchemy_utils/types/pg_composite.py#L298 where `composite.name` is not quoted. In PostgreSQL, types and column names must be quoted in order to preserve case.

Thus, in order to support a mixed case `CompositeType` , I propose that the line instead look like:
```python
    return AsIs('(%s)::"%s"' % (', '.join(values), composite.name)) 
```
where `composite.name` is now being quoted after string interpolation.

I'll be submitting a PR soon to fulfill this proposal alongside a new test to back the change.

bamartin125:
Closes #583

### DIFF
--- a/sqlalchemy_utils/types/pg_composite.py
+++ b/sqlalchemy_utils/types/pg_composite.py
@@ -184,13 +184,16 @@ class CompositeType(UserDefinedType, SchemaType):
 
             return CompositeElement(self.expr, key, type_)
 
-    def __init__(self, name, columns):
+    def __init__(self, name, columns, quote=None):
         if psycopg2 is None:
             raise ImproperlyConfigured(
                 "'psycopg2' package is required in order to use CompositeType."
             )
-        SchemaType.__init__(self)
-        self.name = name
+        SchemaType.__init__(
+            self,
+            name=name,
+            quote=quote
+        )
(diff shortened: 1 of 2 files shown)


bamartin125:
Changes were made to this PR to instead utilize SQLAlchemy's identifier quoting rules/options.

bamartin125:
Well, I had claimed that all tests had passed on my system. The automated test results made me take another look and I obviously found the linting error (newline) I missed in `test_composite.py`. (now fixed)

I do run into other sqla14 errors on my end as well, but those are not caused by my additions. Another issue will have to solve those errors.

When there are errors in the test results prior to making code changes, it's easy to miss errors that are added. Such is what happened when I added the linting error which I overlooked.

bamartin125:
I've just squashed my commits and rebased with the latest sqlalchemy-utils master commits.

PR #589, opened 2022-04-12
Ignore deprecation warnings caused by backports.zoneinfo

### LINKED ISSUE #590: Ignore ResourceWarnings
The unit tests are currently failing due to a `ResourceWarning`:

```
> warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
E pytest.PytestUnraisableExceptionWarning: Exception ignored in: <socket.socket fd=-1, family=AddressFamily.AF_INET6, type=SocketKind.SOCK_STREAM, proto=6>
E 
E Traceback (most recent call last):
E   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/socket.py", line 331, in makefile
E     buffer = io.BufferedRWPair(raw, raw, buffering)
E ResourceWarning: unclosed <socket.socket fd=11, family=AddressFamily.AF_INET6, type=SocketKind.SOCK_STREAM, proto=6, laddr=('::1', 46770, 0, 0), raddr=('::1', 5432, 0, 0)>
```

The exact location changes from run to run because the warning is being issued by Python's garbage collector (whenever it runs).

## Root cause

The root cause is an unclosed socket in [the pg8000 library](https://github.com/tlocke/pg8000/). During class instantiation, a socket is opened. Under certain circumstances, it's possible for the `.__init__()` method to raise an exception:

https://github.com/tlocke/pg8000/blob/121f754c5d423ea4785c8ed35ee72f5586a56a26/pg8000/core.py#L341-L342

When the exception is raised by the `.__init__()` method, the open socket object becomes inaccessible and cannot be closed except by Python's garbage collector. I've [reported this](https://github.com/tlocke/pg8000/issues/100) and [submitted a PR](https://github.com/tlocke/pg8000/pull/101) to the pg8000 project.

## Mitigation possibilities

This error can be mitigated from within sqlalchemy-utils one of two ways. Option 1 is to change the order of database names in `database_exists()`:

https://github.com/kvesteri/sqlalchemy-utils/blob/197dbfdb24ddd6d91f8194f0e25a53c255058a4f/sqlalchemy_utils/functions/database.py#L484-L490

If `'postgres'` is checked before `database` then this will avoid the bug in pg8000.

Option 2 is to filter this specific `ResourceWarning` in `tox.ini`.

@kvesteri I don't understand the implications of changing the order of the database names in `database_exists()`, so I'm going to avoid touching that code and instead submit a PR that follows option 2.

kurtmckee:
This addresses the warnings that are now escalated to errors during unit testing.

Fixes #590

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -44,3 +44,5 @@ filterwarnings =
     ; Ignore ResourceWarnings caused by unclosed sockets in pg8000.
     ; These are caught and re-raised as pytest.PytestUnraisableExceptionWarnings.
     ignore:Exception ignored:pytest.PytestUnraisableExceptionWarning
+    ; Ignore DeprecationWarnings caused by pg8000.
+    ignore:distutils Version classes are deprecated.:DeprecationWarning
--- a/tox.ini
+++ b/tox.ini
@@ -41,3 +41,6 @@ filterwarnings =
     ; Ignore DeprecationWarnings caused by backports.zoneinfo (Python 3.6).
     ignore:open_text is deprecated. Use files\(\) instead.:DeprecationWarning
     ignore:open_binary is deprecated. Use files\(\) instead.:DeprecationWarning
+    ; Ignore ResourceWarnings caused by unclosed sockets in pg8000.
+    ; These are caught and re-raised as pytest.PytestUnraisableExceptionWarnings.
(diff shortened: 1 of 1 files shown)


kurtmckee:
@kvesteri This PR makes the unit tests green again in CI. :grinning:

Issue #427, opened 2020-03-15
Create a Postgres view in a specific schema


ChiliConSql:
I am trying to create a postgres view in schema "reports". Using

`    
create_view('reports.my_view',...)
`
`
engine.create_all()
`

appears to create the proper view when I check the datbase through psql.
However, trying to query the view with 

`    query(ViewClass).all() 
`

returns a psycopg2 error: **relation "reports.my_view" does not exist.**

The code works perfectly well without specifying a schema.

sachitrajbhandari:
@ChiliConSql, Were you able to specify schema?

nestguti:
I`m working in a solution to add schema as an argument to create_view function

PR #593, opened 2022-04-19
add schema to create_view() and create_materialized_view()

nestguti:

### DIFF
--- a/sqlalchemy_utils/view.py
+++ b/sqlalchemy_utils/view.py
@@ -6,17 +6,21 @@ from sqlalchemy_utils.functions import get_columns
 
 
 class CreateView(DDLElement):
-    def __init__(self, name, selectable, materialized=False):
+    def __init__(self, name, selectable, schema, materialized=False):
         self.name = name
         self.selectable = selectable
         self.materialized = materialized
+        self.materialized = schema
+
 
 
 @compiler.compiles(CreateView)
 def compile_create_materialized_view(element, compiler, **kw):
     return 'CREATE {}VIEW {} AS {}'.format(
         'MATERIALIZED ' if element.materialized else '',
-        compiler.dialect.identifier_preparer.quote(element.name),
+        # compiler.dialect.identifier_preparer.quote(element.name),
+        compiler.dialect.identifier_preparer.format_table(
+            element.name, element.schema, use_schema=True),
         compiler.sql_compiler.process(element.selectable, literal_binds=True),
     )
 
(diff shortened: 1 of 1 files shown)


sqlalchemy_utils/view.py
```diff
@@ -60,7 +65,7 @@ def create_table_from_selectable(
         )
         for c in get_columns(selectable)
     ] + indexes
-    table = sa.Table(name, metadata, *args, **kwargs)
+    table = sa.Table(name, metadata, schema=schema, *args, **kwargs)
```
kurtmckee, line 68: Python will crash with a `SyntaxError` when it encounters this line because all keyword arguments (`schema=schema`) must appear positional arguments (`*args` is the last positional argument in this line).

kurtmckee (changes requested):
@nestguti, thanks for your work on this.

It appears that you didn't run this code before submitting it. Did you mean to open a pull request?

I'm going to close this PR because it is guaranteed to make Python crash.

To improve this PR, please add unit tests that demonstrate what this code is fixing or adding. Please also manually test the code. You can also post to #427 and ask the users in that ticket to test the code in your branch.

sqlalchemy_utils/view.py
nestguti: Hi @kurtmckee, thanks for your fast answer, It is the first time I'm making a PR, so wasn't sure what I should do. I modified the functions because I'm using views and materialized views in an app with flask_sqlalchemy and a postgres DB with several schemas. So I tested those (create_view, create_materialized_view and refresh with and without schema) in my app and they work properly. I can try to modify and add testing functions hopping that does not take much time.

kurtmckee: Great, thank you! Adding tests that demonstrate a bug is fixed -- or demonstrate a feature is working -- is an important part of development. It's particularly important for sqlalchemy-utils because the code owner, kvesteri, requires tests for all code-related PR's.

This particular function, `create_table_from_selectable()`, will cause the crash so please be sure that you're testing this function.

nestguti: Thanks, I already fix that bug, and the function create_table_from_selectable() is used by create_view() and create_materialized_view() so it is already tested

kurtmckee: The current PR contains a syntax error and will always crash. Perhaps you fixed it in a different branch, or fixed it after pushing the commits to the current branch?

Please add unit tests and open a new PR after further testing. Thanks!

PR #597, opened 2022-04-20
427 allows passing schema name in create_view() and create_materialized_view() 

nestguti:
#### add test unit test_views_schema.py and update test_views.py

### DIFF
--- a/tests/test_views_schema.py
+++ b/tests/test_views_schema.py
@@ -8,12 +8,6 @@ from sqlalchemy_utils import (
 )
 
 
-# @pytest.fixture
-# def create_schema(engine):
-#     if not engine.dialect.has_schema(engine, 'main'):
-#         engine.execute(sa.schema.CreateSchema('main'))
-
-
 @pytest.fixture
 def Article(Base, User):
     class Article(Base):


nestguti:
> Hi, reviewing the code I wonder if DropView has support for schema as CreateView. Maybe I'm wrong but it appears it doesn't.

Yes, It does:
`@compiler.compiles(DropView)
def compile_drop_materialized_view(element, compiler, **kw):
    return 'DROP {}VIEW IF EXISTS {} {}'.format(
        'MATERIALIZED ' if element.materialized else '',
        compiler.dialect.identifier_preparer.format_table(
            element.table, use_schema=True),
        'CASCADE' if element.cascade else ''`

kurtmckee:
@nestguti thanks for pointing that out. I edited your comment so it would show up as highlighted Python. In the future you can use this syntax to accomplish the same thing:

````markdown
```python
for code in example:
    print("Use three backticks and the name of the language!")
    print("Then end the code block with another three backticks!")
```
````

nestguti:
> @nestguti thanks for pointing that out. I edited your comment so it would show up as highlighted Python. In the future you can use this syntax to accomplish the same thing:
> 
> ```gfm
> ```python
> for code in example:
>     print("Use three backticks and the name of the language!")
>     print("Then end the code block with another three backticks!")
> ```
> ```

Thanks!

PR #599, opened 2022-04-24
[Feature] add 'create_if_not_exists' for views

eterna2:
### Feature:
Add flag `create_if_not_exists` for `create_view` and `create_materialized_view`.

### Background:
Hi, I am using `create_view` and `create_materialized_view` to generate views via cli. However, i will like the cli to not throw an error if the views already exists. This PR adds a kwargs `create_if_not_exists` for `create_view` and `create_materialized_view`, which will insert `CREATE VIEW {} IF NOT EXISTS` if the flag is set.

The default for the flag is `False` so that everything will be backward compatible.

### DIFF
--- a/sqlalchemy_utils/view.py
+++ b/sqlalchemy_utils/view.py
@@ -6,16 +6,17 @@ from sqlalchemy_utils.functions import get_columns
 
 
 class CreateView(DDLElement):
-    def __init__(self, name, selectable, materialized=False):
+    def __init__(self, name, selectable, materialized=False, create_if_not_exists=False):
         self.name = name
         self.selectable = selectable
         self.materialized = materialized
-
+        self.not_exists = create_if_not_exists
 
 @compiler.compiles(CreateView)
 def compile_create_materialized_view(element, compiler, **kw):
-    return 'CREATE {}VIEW {} AS {}'.format(
+    return 'CREATE {}VIEW {}{} AS {}'.format(
         'MATERIALIZED ' if element.materialized else '',
+        'IF NOT EXISTS ' if element.not_exists else '',
         compiler.dialect.identifier_preparer.quote(element.name),
         compiler.sql_compiler.process(element.selectable, literal_binds=True),
     )
@@ -74,7 +75,8 @@ def create_materialized_view(
     selectable,
     metadata,
     indexes=None,
-    aliases=None
+    aliases=None,
+    create_if_not_exists=False,
 ):
     """ Create a view on a given metadata
 
@@ -87,6 +89,9 @@ def create_materialized_view(
     :param aliases:
         An optional dictionary containing with keys as column names and values
         as column aliases.
+    :param create_if_not_exists:
+        An optional flag to indicate whether to use the 
+        ``CREATE MATERIALIZED VIEW IF NOT EXISTS`` statement.
 
     Same as for ``create_view`` except that a ``CREATE MATERIALIZED VIEW``
     statement is emitted instead of a ``CREATE VIEW``.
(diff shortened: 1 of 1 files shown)


eterna2:
@kurtmckee Not sure if you are the right person. Can u help me have a look at this PR?

kurtmckee:
@eterna2, I don't feel qualified yet to review feature requests; my focus has been ticket triage and bug fixes. The project policy is to have unit tests associated with pull requests, so you'll need to add unit tests that demonstrate that this feature is working.

For example, it will be good to confirm that the function continues to raise an exception when the view already exists (the tests may already check this but you should confirm), and also test that the error is suppressed when `create_if_not_exists` is True.

@kvesteri, would you review this PR? I'm interested in your thoughts on the parameter name `"create_if_not_exists"`. The implementation seems similar to how `"materialized"` is supported, but the parameter name seems inaccurate to my eyes and I don't have alternate suggestions.

sqlalchemy_utils/view.py
```diff
@@ -6,16 +6,17 @@
 
 
 class CreateView(DDLElement):
-    def __init__(self, name, selectable, materialized=False):
+    def __init__(self, name, selectable, materialized=False, create_if_not_exists=False):
```
kvesteri, line 9: Name the parameter `if_not_exists`

sqlalchemy_utils/view.py
```diff
@@ -6,16 +6,17 @@
...
         self.name = name
         self.selectable = selectable
         self.materialized = materialized
-
+        self.not_exists = create_if_not_exists
```
kvesteri, line 13: Use name `if_not_exists`

sqlalchemy_utils/view.py
```diff
@@ -74,7 +75,8 @@ def create_materialized_view(
     selectable,
     metadata,
     indexes=None,
-    aliases=None
+    aliases=None,
+    create_if_not_exists=False,
```
kvesteri, line 79: Use name `if_not_exists`

kvesteri (changes requested):

eterna2:
Hi, I had changed the kwargs to `if_not_exists` as requested. 

Also made several other changes:
- added kwargs `or_replace` for `CREATE [OR REPLACE|ALTER] [MATERIALIZED] VIEW`
- added compilation for `mysql`, `postgres`, `mssql`, and `snowflake` dialects
- updated flake8 rules specifically for view.py to ignore the long docstr and redefinitions for the dialect specific compilation.

I am actually using this for snowflake specifically.

eterna2:
Sorry. Some unnecessarily format changes cuz i was using black to quickly do the autoformatting.

kurtmckee:
Thanks @eterna2! I've kicked off the CI runs. Please also add unit tests that demonstrate the new functionality.

eterna2:
Hi, I have added the unit tests for the new params. I just wrote the test for the generic use case and postgres.

eterna2:
Seem like crypto package is raising a deprecation warning. Do u want me to add a marker to the requirement for py3.6?

kurtmckee:
No, I can add that separately.

sqlalchemy_utils/view.py
```diff
@@ -6,21 +6,149 @@
...
+        'OR REPLACE ' if element.or_replace else '',
+        'MATERIALIZED ' if element.materialized else '',
+        'IF NOT EXISTS ' if element.if_not_exists else '',
+        compiler.dialect.identifier_preparer.quote(element.name),
+        compiler.sql_compiler.process(element.selectable, literal_binds=True),
+    )
+
+
+@compiler.compiles(CreateView, 'postgresql')
+def compile_create_materialized_view_(element, compiler, **kw):
```
kvesteri, line 36: Just name all the dialect specific functions with the dialect prefix. That makes them more readable. So this function would then be `compile_create_materialized_view_postgresql`

kvesteri (changes requested):

sqlalchemy_utils/view.py
eterna2: ok done.

PR #637, opened 2022-09-12
Update color.py for colour-science compatibility

### LINKED ISSUE #636: Optional dependency Colour causes implicit conflict with colour-science
A trivial example to reproduce: 
```
python3 -m venv venv
source venv/bin/activate
pip install colour-science sqlalchemy-utils
python
>>> import sqlalchemy_utils
```

Since this is now an optional dependency, it seems like an additional except to catch AttributeError could be added to [sqlalchemy_utils.types.color](https://github.com/kvesteri/sqlalchemy-utils/blob/master/sqlalchemy_utils/types/color.py) to fix this case.

JayPalm:
This will allow SQLAlchemy-Utils to be compatible with `colour-science` Currently, this import makes this package incompatible with `colour-science` due to conflicting `colour` namespace. Since the `colour` package is already optional, this will allow `sqlalchemy-utils` to be used with `colour-science` without conflict. See [#636](https://github.com/kvesteri/sqlalchemy-utils/issues/636) for the associated issue/discussion.

Closes #636

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,14 @@ Changelog
 Here you can see the full list of changes between each SQLAlchemy-Utils release.
 
 
+Unreleased
+^^^^^^^^^^
+
+- Fix a crash that occurs if the ``colour-science`` package is installed,
+  which shares the same import name as the ``colour`` package that sqlalchemy-utils supports.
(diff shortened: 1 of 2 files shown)


kurtmckee:
Thanks @JayPalm! I've added a changelog entry as well. :+1:

PR #640, opened 2022-10-09
Allow selecting default database when using create_database()

georgepstaylor:
Digital Ocean managed postgres database servers use 'defaultdb' instead of postgres. 
This should be allowed to be changed, rather than hard coded


georgepstaylor:
@kurtmckee, you are correct - apologies didn't catch that! I will resubmit

georgepstaylor:
Hi, is this repo still actively maintained?

PR #643, opened 2022-10-24
Fix version comparison in UUID type

Dinmukhamet:
Recent versions of sqlalchemy have non-digit characters (for example, [here](https://github.com/sqlalchemy/sqlalchemy/blob/8e3ec05b2e4b3f22089b531b737c12c249fd2d8e/lib/sqlalchemy/__init__.py#L260)), which is why this line sometimes breaks.

It is a better practice to use Version class from packaging.version standard library to compare versions.

### DIFF
--- a/sqlalchemy_utils/types/uuid.py
+++ b/sqlalchemy_utils/types/uuid.py
@@ -1,11 +1,12 @@
 import uuid
 
+from packaging.version import Version
 from sqlalchemy import __version__, types, util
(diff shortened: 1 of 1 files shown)


kurtmckee:
@Dinmukhamet thanks for pointing this out! `packaging` is not listed as a dependency of sqlalchemy-utils; it is only installed while the test suite is running.

I would prefer to achieve the same result without introducing `packaging` as a new dependency.

PR #653, opened 2022-10-27
Update setup.py to allow python>3.6 (fix #652)

### LINKED ISSUE #652: sqlalchemy-utils depends on Python 3.6
https://pypi.org/project/SQLAlchemy-Utils/: 

> Requires: Python ~=3.6 

Python3.6 was end-of-life over a year ago, and now Python 3.11 is the newest version.

Our project also includes the very popular `requests` library, which only supports Python 3.7 and up.

There is no way to override this in `poetry`, the common Python installer.

https://stackoverflow.com/questions/60567732/poetry-force-install-when-versions-are-incompatible

rec:


kurtmckee:
As I wrote in #652, the `~=` syntax follows the [compatible release](https://peps.python.org/pep-0440/#compatible-release) syntax. It is interpreted to mean `>= 3.6, == 3.*`.

Changing this to just `>=3.6` would indicate that this is compatible even with a hypothetical Python 4 release, which is not a desirable outcome.

tucked (commented):
IMO, this is a good idea since this is a library and should be as flexible as possible, only blocking dependencies that are _known_ to not work. Consider this table:
| | Compatible | Incompatible |
| --- | ---| --- |
| `~=` | Users cannot try running with 4.x, even though it would work fine. | Users cannot try running with 4.x, but it probably wouldn't work anyways (though, it might because they may not use incompatible features). |
| `>=` | Users can run with 4.x seamlessly. | Users can try running with 4.x, but it may not work. |

The bottom left and top right show the best case scenario for each operator, while the top left and the bottom right show the worst case, and we can see `>=` outperforming in both cases, primarily because it leaves the user/dependent/downstream in control of their own destiny.

kurtmckee:
IMO, this is not a good idea since semantic versioning specifically allows breaking changes between major releases, and because Python itself made such radical changes between the last major releases (2 to 3) that it fractured codebases and the community for an entire decade.

When Python 4.x is released, sqlalchemy-utils will publish a release that claims support for it at that time.

PR #657, opened 2022-11-15
add changelog entry in 0.38.3 about dropped Pyhton 3.4/3.5

fmigneault:
The change was introduced in 0.38.3 (in https://github.com/kvesteri/sqlalchemy-utils/commit/51464b68342c0462aa833ad507e91f0024e45293) by dropping Python 3.4/3.5 support while adding 3.10, but it wasn't reported.


kurtmckee:
Thanks for catching this, @fmigneault!

PR #668, opened 2022-12-23
Can we add comparison functions to Ltrees? 

salimfadhley:
The rationale for this - I use a lot of Ltrees. It would be nice to be able to sort them. That requires that Ltrees can be compared to other Ltrees.


salimfadhley:
Any chance we could have this merged?

kvesteri (approved):

PR #701, opened 2023-04-07
Add psycopg3 support for database functions

LerikP:
Sqlalchemy now supports psycopg3, that is called simply psycopg and create(delete)_database functions do not detect this dialect properly


kurtmckee:
It doesn't appear that this is installing psycopg3 so it's available in the test suite.

neumann-nico:
@LerikP Thanks for the pull request, it works nicely :)

PR #705, opened 2023-05-22
i18n: Support SQLAlchemy 2 MappedColumn

madsodgaard:
When trying to use the internalization module with SQLAlchemy and the `mapped_column` style, you run into the following error:
`AttributeError: 'MappedColumn' object has no attribute 'key'`.

This adds a workout around by accessing the key though `attr.column.key` if the attribute is a `MappedColumn`. Not sure if the implementation is the best approach in terms of providing compatibility with 1.X, so any input would be great!

### DIFF
--- a/sqlalchemy_utils/i18n.py
+++ b/sqlalchemy_utils/i18n.py
@@ -11,6 +11,11 @@ try:
 except ImportError:
     babel = None
 
+try:
+    from sqlalchemy.orm import MappedColumn
+except ImportError:
+    MappedColumn = None
+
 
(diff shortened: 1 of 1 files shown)


atugushev-aghanim (commented):
Thanks! Just bumped into this PR after [implementing](https://github.com/atugushev/sqlalchemy-utils/commit/801183a73bf4a84b54feaefc62b06ab3fc9eb815) this by myself.

I could help with the tests if you don't mind to increase the chances of this being merged. What do you think?

PR #707, opened 2023-06-01
Fix bug preventing use of schema prefix in name of CompositeType

ryan-pypestream:
Fixed an error which occurred when a CompositeType has a schema prefix.


kurtmckee:
Thanks for working to fix this, Ryan!

Would you add a test to the suite that demonstrates that this bug exists and is fixed by this change?

ryan-pypestream:
Sure, do you have documentation somewhere on how to run the tests locally?

kurtmckee:
It looks like you might need to add `AccountWithSchema` to the `init_models()` fixture in the test file. Currently the tests are failing because `account_with_schema` doesn't exist.

PR #710, opened 2023-06-09
[view] Support CREATE OR REPLACE

mgorven:


kurtmckee (commented):
@mgorven Thanks for submitting this!

Would you add some tests that exercise this feature? It looks like these tests will be worthwhile:

* `ValueError` is raised when `materialized` and `replace` are both true
* A view is created when `replace` is true and _it doesn't exist_
* A view is replaced when `replace` is true and _it already exists_

mgorven:
Sure. Are there any docs on how to run tests locally? It seems like they require a real database.

mgorven:
As was discovered in #621, the tests don't actually create views (they create a table). I've changed the tests to run `metadata.create_all()` instead, which should create the view and backing tables.

mgorven:
Only test failures are `TestAggregatesWithManyToManyRelationships`, which are failing in master.

sqlalchemy_utils/view.py
```diff
@@ -124,7 +128,8 @@ def create_view(
     name,
     selectable,
     metadata,
-    cascade_on_drop=True
+    cascade_on_drop=True,
+    replace=False,
```
kurtmckee, line 132: Apologies, I just realized that the *replace* parameter needs to be documented in the function docstring. This will be picked up the next time the documentation is generated.

I think the parameter documentation can be added below `:param metadata:`. Perhaps something like:

```text
    :param replace:
        If ``True``, any view with the same *name* will be replaced.
        Note that materialized views cannot be replaced.
```

But please don't be limited by that suggestion!

kurtmckee (changes requested):
One last thing that I finally realized: the *replace* parameter needs a bit of documentation in the `create_view()` docstring.

It'll be rendered automatically the next time the docs are built, and will appear like the other parameters in [the `create_view()` documentation](https://sqlalchemy-utils.readthedocs.io/en/latest/view.html#create-view).

kurtmckee (approved):

kurtmckee:
Thanks for your work on this, @mgorven!

mariusfeldmann:
Hey, thanks for the work on this!

Is there a reason why the replace parameter is missing on the `create_materialized_view` method?

Issue #719, opened 2023-10-13
generic_relationship breaks with SqlAlchemy 2.0.22


holyoaks:
I have two table classes defined using SqlAlchemy 2.0 syntax. Yesterday, using SqlAlchemy 2.0.21, my application worked fine. This morning, using SqlAlchemy 2.0.22, my application will not start. So why am I creating an issue here instead of SqlAlchemy?

Of the two table classes, one has a `generic_relationship` defined, while the other does not. Stepping through the code as the application attempts to start, the table class _without_ a `generic_relationship` is mapped correctly, no issues. The second table class, _with_ a `generic_relationship`, raises:

`TypeError: AttributeImpl.__init__() missing 1 required positional argument: 'dispatch'`

The two table classes, along with the definition of the `Base` declarative class, are shown below:

```python
from sqlalchemy import Boolean, DateTime, ForeignKey
from sqlalchemy.dialects.postgresql import BIGINT, JSONB
from sqlalchemy.ext.mutable import MutableDict
from sqlalchemy.orm import DeclarativeBase, mapped_column, relationship
from sqlalchemy.sql import func
from sqlalchemy.sql.schema import Index
from sqlalchemy.sql.sqltypes import Integer, String, Unicode
from sqlalchemy_utils import generic_relationship


class Base(DeclarativeBase):
    pass


class EventCategory(Base):
    __tablename__ = "eventcategory"
    id = mapped_column(BIGINT, primary_key=True, index=True)
    icon_url = mapped_column(
        String, nullable=True
    )
    is_archived = mapped_column(Boolean, default=False)
    label = mapped_column(String)

    events = relationship("Event", back_populates="category")


class Event(Base):
    __tablename__ = "event"
    id = mapped_column(BIGINT, primary_key=True, index=True)
    action = mapped_column(String)
    category_id = mapped_column(BIGINT, ForeignKey("eventcategory.id", ondelete="CASCADE"))
    datetime_occurred = mapped_column(DateTime(timezone=True), default=func.now())
    is_internal = mapped_column(Boolean, default=False)
    object_id = mapped_column(Integer, nullable=True)
    object_data = mapped_column(MutableDict.as_mutable(JSONB), default={})
    object_type = mapped_column(Unicode)
    permalink = mapped_column(String, nullable=True)
    source_user_name = mapped_column(String)
    text_map_values = mapped_column(MutableDict.as_mutable(JSONB), default={})

[...]

PR #725, opened 2024-01-18
Update GenericAttributeImpl to work with SqlAlchemy 2.0.22

danigm:
This change fixes the issue, but I'm not sure if this should be fixed in SQLAlchemy directly.

I've been debugging and found the problem in this call where there's a missing parameter. Looks like the `AttributeImpl` class now requires [4 positional parameters](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/orm/attributes.py#L838), and the [register_attribute mapping is not calling this](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/orm/attributes.py#L2620-L2622) correctly.

Fix https://github.com/kvesteri/sqlalchemy-utils/issues/719

### DIFF
--- a/sqlalchemy_utils/generic.py
+++ b/sqlalchemy_utils/generic.py
@@ -13,6 +13,13 @@ from .functions.orm import _get_class_registry
 
 
 class GenericAttributeImpl(attributes.ScalarAttributeImpl):
+    def __init__(self, *args, **kwargs):
+        # arguments received (class, key, dispatch)
+        # The attributes.AttributeImpl requires (class, key, default_function, dispatch)
+        # Setting None as default_function here
+        args = args[:2] + (None, ) + args[2:]
+        super().__init__(*args, **kwargs)
+
     def get(self, state, dict_, passive=attributes.PASSIVE_OFF):
         if self.key in dict_:
             return dict_[self.key]


josecsotomorales:
Reported this issue https://github.com/kvesteri/sqlalchemy-continuum/issues/339 and ended up here, what's needed to merge this PR?

josecsotomorales:
@kurtmckee can you take a look at this?

stabacco:
I'm also interested in this to be merged. Currently I have to pin the sqlalchemy version to 2.0.21

kurtmckee:
Test suite appears to be failing for unrelated reasons.

@danigm Thanks for contributing this fix!

josecsotomorales:
@kurtmckee @kvesteri happy to be a contributor to this project to get this released

marksteward:
Does this fix retain compatibility with 2.0.21?

marksteward:
I've tested this and it breaks support with sqla < 2.0.21, including 1.4.

`register_attribute_impl` is actually called [a few lines above](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/orm/attributes.py#L2614), which has a different signature. Previously [they all took 4 positional arguments](https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_4/lib/sqlalchemy/orm/attributes.py#L2150-L2159).

So to avoid breaking anyone who doesn't immediately update to 2.0.21, I think this code needs to count the number of args, and only add the extra `callable_` if needed.

josecsotomorales:
@marksteward I haven't looked at this yet, this was merged to master tough, I will try to get tests to at least pass with the current master version in this [PR](https://github.com/kvesteri/sqlalchemy-utils/pull/732) first, then come back to this, if you can help with a PR

PR #755, opened 2024-09-06
fix(pendulum): Make PendulumDateTime work with postgres

### LINKED ISSUE #754: pendulum datetime fails to preserve the timezone
I‘ve just tried out the pendulum datetime type and found that it doesn’t deal with timezones correctly.

Test case:

```python
class Event(db.Model):
    created_at = orm.mapped_column(sqlalchemy_utils.EnrichedDateTime())


date_str = "2015-06-22T12:20:29.554Z"
event = Event(created_at=pendulum.parse(date_str))
db.session.add(event)
db.session.commit()
assert event.created_at == pendulum.parse(date_str)
# Fails: event.created_at was shifted to my local timezone
```

I think this is due to `PendulumDateTime.process_result_value()` interpreting the date from the DB (without timezone) as being in the local timezone instead of UTC.

torotil:
fixes #754

### DIFF
--- a/sqlalchemy_utils/types/enriched_datetime/pendulum_datetime.py
+++ b/sqlalchemy_utils/types/enriched_datetime/pendulum_datetime.py
@@ -37,5 +37,7 @@ class PendulumDateTime:
 
     def process_result_value(self, impl, value, dialect):
         if value:
-            return pendulum.DateTime.instance(value.replace(tzinfo=datetime.timezone.utc))
+            return pendulum.DateTime.instance(
+                value.replace(tzinfo=datetime.timezone.utc)
+            )
         return value
--- a/sqlalchemy_utils/types/enriched_datetime/pendulum_datetime.py
+++ b/sqlalchemy_utils/types/enriched_datetime/pendulum_datetime.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+import datetime
 
 from ...exceptions import ImproperlyConfigured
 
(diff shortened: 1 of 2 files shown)


sqlalchemy_utils/types/enriched_datetime/pendulum_datetime.py
```diff
@@ -24,26 +24,18 @@ def _coerce(self, impl, value):
...
     def process_bind_param(self, impl, value, dialect):
         if value:
-            return self._coerce(impl, value).in_tz('UTC')
+            return self._coerce(impl, value).in_tz('UTC').naive()
         return value
 
     def process_result_value(self, impl, value, dialect):
         if value:
-            return pendulum.parse(value.isoformat())
+            return pendulum.DateTime.instance(value.replace(tzinfo=datetime.timezone.utc))
```
kurtmckee, line 40: flake8 is unhappy. Gonna commit this suggestion to try making it happy.

```suggestion
            return pendulum.DateTime.instance(
                value.replace(tzinfo=datetime.timezone.utc)
            )
```

kurtmckee (changes requested):

kurtmckee (approved):

kurtmckee:
Thanks for this fix, and for including a test case to demonstrate it!

torotil:
… and thanks to you for merging this so quickly!

torotil:
@kurtmckee I have now taken a more in-depth look at the code in enriched_datetime. To me it seems a bit overly complicated. I think the indirection using the EnrichedDate / EnrichedDateTime objects is not actually worth it. There is  almost no code shared between those classes. I have made a draft of removing that (and updating the the docs). Are you interested in getting a PR in that direction? (The draft can be found here: https://github.com/kvesteri/sqlalchemy-utils/compare/master...torotil:sqlalchemy-utils:refactor-enriched_datetime)

I also discovered that the arrow implementation already handles both timezone-aware and naive objects while the pendulum implementation seems to only handle naive ones (especially after the changes in this PR). I suppose this would need more test cases on the pendulum side.

kurtmckee:
PRs are welcome, but I hope that there will be others with more knowledge of the codebase that @kvesteri is able to bring on board for maintenance.

PR #758, opened 2024-09-12
Fix: TranslationHybrid setter_factory not persisting changes

rmcilraith:
Marks the attribute in the setter factory as dirty so SQLAlchemy includes the changes in the next commit.

Currently without this, running `session.commit()` followed by `session.refresh(model)` does not persist changes to the model translation hybrid field updated.

Tested with SQLAlchemy version 1.4.37


rmcilraith:
The issue was due to the field being updated by translation hybrid being of type `JSONB` and was fixed by using `MutableDict` from SQLAlchemy.

PR #765, opened 2024-11-26
Fix arrow python type

tim-haselhoff:
Fix sqlalchemy returning datetime in python instead of arrow when using arrow

### DIFF
--- a/sqlalchemy_utils/types/arrow.py
+++ b/sqlalchemy_utils/types/arrow.py
@@ -1,3 +1,4 @@
+from arrow import Arrow
 from ..exceptions import ImproperlyConfigured
 from .enriched_datetime import ArrowDateTime
 from .enriched_datetime.enriched_datetime_type import EnrichedDateTimeType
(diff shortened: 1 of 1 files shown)


benedikt-bartscher (approved):

benedikt-bartscher:
@kvesteri foremost: thanks for maintaining sqlalchemy-utils. We really enjoyed using it so far.

Is there any way we can get this in? I would love to help if there is anything left to do

torotil:
I think that this whole part of sqlalchemy-utils needs some work (see #761) … and it seems that there is currently noone around to maintain this code properly (ie. with an opinion on its architecture).

Having said that I had luck getting a bug fix in (#754) by also providing a test case that demonstrates the issue.

benedikt-bartscher:
We could add a small test for it if required. @tim-haselhoff wdyt?

sqlalchemy_utils/types/arrow.py
```diff
@@ -1,3 +1,4 @@
+from arrow import Arrow
```
kurtmckee, line 1: This results in an `ImportError` if arrow isn't installed; the current architecture is to allow for imports within the `sqlalchemy_utils` package but only raise an exception if a class is used but the underlying dependency isn't installed.

kurtmckee (changes requested):
There appear to be several unnecessary changes here; the only thing that appears necessary is to add:

```python
        @property
        def python_type(self):
            return Arrow
```

Please add a unit test to demonstrate that this is resolving a bug.

kurtmckee:
Closing for lack of response.

PR #776, opened 2025-05-06
Upgrading msodbcsql to msodbcsql18

alouafi:
# Upgrade MSSQL driver from ODBC 17 to ODBC 18 with connection handling improvements

This PR upgrades the Microsoft SQL Server driver from ODBC Driver 17 to ODBC Driver 18, making several necessary adjustments to ensure proper functionality:

## Changes
- Updated the installation script to use `msodbcsql18` instead of `msodbcsql17`
- Changed the default MSSQL driver in the test configuration to "ODBC Driver 18 for SQL Server"
- Added `TrustServerCertificate=yes` parameter to the connection string to address SSL certificate validation issues with the newer driver
- Streamlined database operations by using `isolation_level='AUTOCOMMIT'` consistently
- Simplified the database cleanup code in MSSQL by removing redundant logic while maintaining proper connection handling
- Removing piece of code that was cleaning running transactions because AUTOCOMMIT was not set to True in `database_exists`

## Additional Updates
- Removed Ubuntu 20.04 test configuration for Python 3.9, as GitHub is retiring Ubuntu 20.04 runners on April 15, 2025 (see https://github.com/actions/runner-images/issues/11101)
- Added specific handling for MSSQL in the `database_exists` function, improving reliability

These changes ensure compatibility with current Microsoft SQL Server drivers while maintaining support across all tested Python versions and SQLAlchemy versions.

### DIFF
--- a/sqlalchemy_utils/functions/database.py
+++ b/sqlalchemy_utils/functions/database.py
@@ -485,7 +485,7 @@ def database_exists(url):
             text = "SELECT 1 FROM pg_database WHERE datname='%s'" % database
             for db in (database, 'postgres', 'template1', 'template0', None):
                 url = _set_url_database(url, database=db)
-                engine = sa.create_engine(url)
+                engine = sa.create_engine(url, isolation_level='AUTOCOMMIT')
                 try:
                     return bool(_get_scalar_result(engine, sa.text(text)))
                 except (ProgrammingError, OperationalError):
@@ -508,6 +508,14 @@ def database_exists(url):
                 # The default SQLAlchemy database is in memory, and :memory: is
                 # not required, thus we should support that use case.
                 return True
+        elif dialect_name == 'mssql':
+            text = "SELECT 1 FROM sys.databases WHERE name = '%s'" % database
+            url = _set_url_database(url, database='master')
+            engine = sa.create_engine(url, isolation_level='AUTOCOMMIT')
+            try:
+                return bool(_get_scalar_result(engine, sa.text(text)))
+            except (ProgrammingError, OperationalError):
+                return False
         else:
             text = 'SELECT 1'
             try:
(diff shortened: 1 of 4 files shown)


sqlalchemy_utils/functions/database.py
```diff
@@ -655,6 +655,18 @@ def drop_database(url):
...
             text = f'DROP DATABASE {quote(conn, database)}'
             conn.execute(sa.text(text))
+    elif dialect_name == 'mssql':
+        with engine.begin() as conn:
+            # Set the database to single user mode to disconnect all users
+            text = f'''
+            ALTER DATABASE {quote(conn, database)}
+            SET SINGLE_USER WITH ROLLBACK IMMEDIATE;
+            '''
+            conn.execute(sa.text(text))
```
alouafi, line 665: similarly to what is being done above for postgresql

marksteward: Hmm, I don't see why this should suddenly be needed. If the DB's still in use, that suggests something's still connected to it, and we should work out why that's happening.

alouafi:
Hi @marksteward I just pushed some changes on my branch (cherry-picked some of your changes as well 🙏 see here: https://github.com/alouafi/sqlalchemy-utils), and it seems like the CI is successfully running, let me know what you think about it!

sqlalchemy_utils/functions/database.py
marksteward: The postgres code was introduced to fix "bugs found in use of psycopg2" (https://github.com/kvesteri/sqlalchemy-utils/commit/5d741c16662cde063e6930e7d3fe87b00ae9b8bb) which also sounds suspect

alouafi: I tried to debug a bit by printing a bunch of info on the idle transaction and it seems like the transaction is in state 2, which indicates it's active but not currently executing a command + the system couldn't find any recent SQL query tied to this transaction (so I guess it's been idle for a while or has no query). So there's definitely a transaction being open somewhere in the code that's not rollbacked nor commited, but it's hard to pinpoint exactly where. Still don't understand why this problem pops up just now. I'll try to investigate more if I get the time (otherwise I'll wait for your expertise 🤓 )

alouafi: Hey Mark, actually I managed to pinpoint the issue! The test that was failing was running sequentially:

1. create_database
2. database_exists
3. drop_database

However, `database_exists` does not commit its query (but `create_database` and `drop_database` do). So I added autocommit to both mssql and postgresql in `database_exists`, and now the tests are running just fine. I also removed the piece of code that was cleaning the idle transactions postgresql and standardized the way autocommit is used a bit. All checks are running fine: https://github.com/alouafi/sqlalchemy-utils/tree/patch-1

Let me know what you think about it!

marksteward: Amazing! I'll give that a go this evening.

marksteward: Thanks, this looks great. I've squashed your commits and edited the message as e2cbc08, hope that's OK

marksteward:
Rebased and merged as a4d8b5b and e2cbc08

marksteward:
Thanks for all your help with this!

PR #781, opened 2025-07-10
Migrated to PEP 621 and added Python 3.13 support

josecsotomorales:


pyproject.toml
```diff
@@ -0,0 +1,95 @@
+[build-system]
+requires = ["setuptools>=45", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "SQLAlchemy-Utils"
+version = "0.41.2"
+description = "Various utility functions for SQLAlchemy."
+readme = "README.rst"
+license = {text = "BSD"}
```
kurtmckee, line 10: When building the package, warnings are emitted to switch to simple SPDX license identifiers:

```suggestion
license = "BSD-3-Clause"
```

This was tested using:

```
$ pip install build
$ python -m build

        ********************************************************************************
        Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

        By 2026-Feb-18, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.

        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
        ********************************************************************************
```

kurtmckee (changes requested):
I found some warnings when building the package and have noted the changes needed to resolve those warnings.

Thanks for your work on this! :partying_face:

kurtmckee:
Looks great! Thanks for your work on this! :rocket:

PR #793, opened 2025-09-30
Fix AttributeError with Sequence defaults in instant_defaults_listener

wadinj:
The instant_defaults_listener was raising AttributeError when encountering columns with Sequence defaults, as it attempted to access the 'arg' attribute which doesn't exist on Sequence objects (only on ColumnDefault objects).

This occurred when using force_instant_defaults() with models that have database sequences for primary keys or other columns, causing failures in scenarios like Celery task result storage.

Resolution:
- Add hasattr check for 'arg' attribute before accessing it
- This generic approach handles Sequence and any other default types that don't have an 'arg' attribute
- Add test case covering Sequence defaults to prevent regression


kurtmckee:
Thanks for this, and for adding a test case!

wadinj:
Any change to get a proper version with this?

PR #795, opened 2025-10-04
Always dispose of engines in create_database and drop_database

ashanbrown:
To ensure timely cleanup of resources, dispose of any engines after using them.  

This need surfaced because psyopg3 reports a warning later if a connection if gc'd, making it desirable to dispose of the engine (and its connection pool) even if the command fails. 

This pattern is already in place for `database_exists`, so I'm just suggesting we extend it for `create_engine` and `drop_database`.

### DIFF
--- a/tests/functions/test_database.py
+++ b/tests/functions/test_database.py
@@ -177,7 +177,7 @@ def test_create_engine(sqlite_memory_dsn):
 
 
 def test_create_engine_always_disposes(sqlite_memory_dsn):
-    """Test that engine creation context manager still dispoes of an engine when an exception is raised."""
+    """Test that engine creation context manager still disposes of an engine when an exception is raised."""
     with pytest.raises(RuntimeError, match='it failed'):
         with _create_engine(sqlite_memory_dsn) as engine:
             pool = engine.pool


sqlalchemy_utils/functions/database.py
```diff
@@ -562,35 +562,37 @@ def create_database(url, encoding='utf8', template=None):
...
-
-        with engine.begin() as conn:
-            text = "CREATE DATABASE {} ENCODING '{}' TEMPLATE {}".format(
-                quote(conn, database), encoding, quote(conn, template)
-            )
-            conn.execute(sa.text(text))
+    try:
+        if dialect_name == 'postgresql':
+            if not template:
+                template = 'template1'
```
kurtmckee, line 568: This kicks off a `try/except` block that is larger than it should be.

I recommend modifying this so that the conditionals exclusively determine what the `text` variable should be.

```python
text = ''
if dialect_name == 'postgresql':
    if not template:
        template = 'template1'
    text = "CREATE DATABASE {} CHARACTER SET = '{}'".format(
        quote(conn, database), encoding
    )
elif dialect_name == 'mysql':
    ...
    text = ...
elif dialect_name == 'sqlite' and database != ':memory:':
    if database:
        text = 'CREATE TABLE DB(id int); DROP TABLE DB;'
else:
    text = f'CREATE DATABASE {quote(conn, database)}'
```

and then reduce the `try/except` block to something like this, which accommodates the possibility that `text` is an empty string.

```python
if text:
    try:
        with engine.begin() as conn:
            conn.execute(sa.text(text))
    finally:
        engine.dispose()
```

(It's possible that `engine.dispose()` still needs to be run, so please check whether the `if text:` conditional should actually be inside the `try` block.)

kurtmckee (changes requested):
@ashanbrown Thanks for submitting this.

Please add a test that demonstrates that this issue both exists _and_ is fixed by this change. Thanks!

sqlalchemy_utils/functions/database.py
ashanbrown: At least one of the examples (sqlite), requires multiple executing multiple `text` values, so this could possibly get a little messy.  If we just had a context manager provide an engine that is always disposed, would that resolve your concern?  Then we could also clean up duplicate code between create_database and drop_database.

tests/functions/test_database.py
```diff
@@ -163,3 +164,11 @@ class TestDatabaseMssql(DatabaseTest):
...
         pytest.importorskip('pyodbc')
         return 'db_test_sqlalchemy_util'
+
+
+def test_create_engine(sqlite_memory_dsn):
+    with _create_engine(sqlite_memory_dsn) as engine:
+        pool = engine.pool
+
+    # a disposed engine should not have the same pool
+    assert engine.pool is not pool
```
ashanbrown, line 174: We could also mock the dispose method if this doesn't seem sufficient.

sqlalchemy_utils/functions/database.py
kurtmckee: I think I addressed that by putting both commands into the same line. I don't anticipate it being messy, but could be wrong.

```python
elif dialect_name == 'sqlite' and database != ':memory:':
    if database:
        text = 'CREATE TABLE DB(id int); DROP TABLE DB;'
```

Issue #803, opened 2026-01-30
SQLA2.1: compatibility with changes to ScalarAttributeImpl


Dev-iL:
Hello,

I [recently attempted](https://github.com/apache/airflow/pull/61229) to test the compatibility of Apache Airflow with the SQLAlchemy 2.1 beta. CI reported multiple issues related to sqlalchemy-utils attempting to import `sqlalchemy.orm.attributes.ScalarAttributeImpl`, e.g.

```none
from sqlalchemy_utils import UUIDType
/usr/python/lib/python3.13/site-packages/sqlalchemy_utils/__init__.py:46: in <module>
    from .generic import generic_relationship  # noqa
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/usr/python/lib/python3.13/site-packages/sqlalchemy_utils/generic.py:15: in <module>
    class GenericAttributeImpl(attributes.ScalarAttributeImpl):
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E   AttributeError: module 'sqlalchemy.orm.attributes' has no attribute 'ScalarAttributeImpl'
```
See more examples in the CI logs below:
- https://github.com/apache/airflow/actions/runs/21493207091/job/61923032136
- https://github.com/apache/airflow/actions/runs/21493207091/job/61923032069

I believe that the breaking change was this:
- https://github.com/sqlalchemy/sqlalchemy/commit/f7e2b1a6e5a2188ecc6937a565b93937581a2d97

Dev-iL:
Closing as duplicate of https://github.com/kvesteri/sqlalchemy-utils/issues/800

Issue #800, opened 2025-12-11
AttributeError: module 'sqlalchemy.orm.attributes' has no attribute 'ScalarAttributeImpl'


ericbuehl:
Using sqlalchemy 2.1 beta this breaks as a result of https://github.com/sqlalchemy/sqlalchemy/issues/10497

amotl:
Hi. We are hitting this as well. Do you have an idea how to resolve it?

-- https://github.com/crate/cratedb-examples/actions/runs/26726386632/job/78762276700?pr=1749#step:7:696

PR #804, opened 2026-01-30
Workaround for importing `ScalarAttributeImpl` on SQLA2.1

Dev-iL:
Handle the below by disabling the `GenericAttributeImpl` class on SQLA2.1:
```none
.../sqlalchemy_utils/generic.py:15: in <module>
    class GenericAttributeImpl(attributes.ScalarAttributeImpl):
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E   AttributeError: module 'sqlalchemy.orm.attributes' has no attribute 'ScalarAttributeImpl'
```

Note: although it looks like a lot has changed, this PR only adds ~10 lines and the rest are indentation changes.

Related #800, #803

### DIFF
--- a/sqlalchemy_utils/generic.py
+++ b/sqlalchemy_utils/generic.py
@@ -11,95 +11,108 @@ from .exceptions import ImproperlyConfigured
 from .functions import identity
 from .functions.orm import _get_class_registry
 
+try:
+    BaseScalarImpl = attributes.ScalarAttributeImpl
+except AttributeError:
+    BaseScalarImpl = None
 
-class GenericAttributeImpl(attributes.ScalarAttributeImpl):
-    def __init__(self, *args, **kwargs):
-        """
-        The constructor of attributes.AttributeImpl changed in SQLAlchemy 2.0.22,
-        adding a 'default_function' required positional argument before 'dispatch'.
-        This adjustment ensures compatibility across versions by inserting None for
-        'default_function' in versions >= 2.0.22.
+if BaseScalarImpl is None:
 
-        Arguments received: (class, key, dispatch)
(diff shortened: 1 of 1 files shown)


kurtmckee:
@Dev-iL It looks like sqlalchemy has [an example](https://docs.sqlalchemy.org/en/20/_modules/examples/generic_associations/generic_fk.html) that we should probably be following (and that base example appears to have existed in sqlalchemy 1.4 as well).

I think that instead of causing this to fail on new versions of sqlalchemy, this PR should instead migrate the code to incorporate the example that sqlalchemy has defined.

[From the maintainers of sqlalchemy](https://github.com/sqlalchemy/sqlalchemy/discussions/10474#discussioncomment-7316373) --

> Casey: apparently they define a so called generic_relationship. I don't know what it does

> zzzeek: it's probably what we have here: https://docs.sqlalchemy.org/en/20/_modules/examples/generic_associations/generic_fk.html
>
> but I really would have preferred they didnt make their own attribute impl

I'm closing this PR because it causes an outright failure on new versions of sqlalchemy instead of migrating the code to supported usage.

Dev-iL:
> I'm closing this PR because it causes an outright failure on new versions of sqlalchemy instead of migrating the code to supported usage.

The failure is already present. The suggested workaround turns the **import-time** error, into a **runtime** error. In other words, on SQLA2.1 sqlalchemy-utils can't even be imported at the moment since the broken import happens at the top level `__init__.py`. With the suggested modification, if the user doesn't need the disabled functionality at least they can use everything else.

I think it is a suitable solution as a temporary measure until the "proper" way that you mentioned is implemented. Of course it wouldn't hurt to mention this behavior explicitly in the docs.

PR #813, opened 2026-07-07
Fix ChoiceType returning raw scalar for falsy codes (0, empty string)

gaoflow:
The list-of-tuples `ChoiceType` fails to coerce a valid but *falsy* stored code back into a `Choice`, while the `Enum` variant handles it correctly — the two sibling implementations disagree on their shared contract.

```python
import sqlalchemy as sa
from sqlalchemy_utils.types.choice import ChoiceType

t = ChoiceType([(0, "inactive"), (1, "active")], impl=sa.Integer())
t.process_result_value(0, None)   # -> 0        (raw int, WRONG)
t.process_result_value(1, None)   # -> <Choice code=1 ...>   (correct)
```

So reading a row whose stored code is `0` (a very common pattern for integer choice lists) or `""` (empty-string code) yields the raw scalar instead of a `Choice`, e.g. `item.status` comes back as `0` rather than `Choice(0, "inactive")`.

**Cause.** `ChoiceTypeImpl.process_result_value` tests `if value:` (truthiness), which is false for `0`/`""`. The `Enum` sibling `EnumTypeImpl.process_result_value` guards with `is None` (via `_coerce`), and `_coerce` itself uses `is None` — only this implementation used truthiness.

**Fix.** Guard with `is None`, matching the sibling and `_coerce`. `process_bind_param` needs no change (a `Choice` is truthy and raw codes pass through unchanged).

**Tests** (`tests/types/test_choice.py`, `TestChoiceTypeFalsyCode`): the falsy int code `0` and empty-string code now wrap into a `Choice`; an explicit parity assertion that the tuple-list and `Enum` variants agree on code `0`; and a full SQLite round-trip asserting `item.status` reads back as `Choice(0, "inactive")`. `tests/types/test_choice.py`: 21 passed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,11 @@ Changelog
 
 Here you can see the full list of changes between each SQLAlchemy-Utils release.
 
+Unreleased
+^^^^^^^^^^
+
+- Fix ``ChoiceType`` returning the raw scalar instead of a ``Choice`` for falsy codes such as ``0`` or the empty string (#813)
+
 0.42.1 (2025-12-12)
 ^^^^^^^^^^^^^^^^^^^
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,9 @@ Here you can see the full list of changes between each SQLAlchemy-Utils release.
 Unreleased
 ^^^^^^^^^^
 
-- Fix ``ChoiceType`` returning the raw scalar instead of a ``Choice`` for falsy codes such as ``0`` or the empty string (#813)
+- Fix ``ChoiceType`` returning the raw scalar instead of a ``Choice`` for falsy codes such as ``0`` or the empty string. (#813)
+
+  NULL values continue to return ``None``.
 
 0.42.1 (2025-12-12)
 ^^^^^^^^^^^^^^^^^^^
--- a/sqlalchemy_utils/types/choice.py
+++ b/sqlalchemy_utils/types/choice.py
@@ -183,9 +183,9 @@ class ChoiceTypeImpl:
         return value
 
     def process_result_value(self, value, dialect):
-        if value is not None:
-            return Choice(value, self.choices_dict[value])
-        return value
+        if value is None:
+            return None
+        return Choice(value, self.choices_dict[value])
 
 
 class EnumTypeImpl:
--- a/sqlalchemy_utils/types/choice.py
+++ b/sqlalchemy_utils/types/choice.py
@@ -183,7 +183,7 @@ class ChoiceTypeImpl:
         return value
 
     def process_result_value(self, value, dialect):
-        if value:
+        if value is not None:
             return Choice(value, self.choices_dict[value])
         return value
 
--- a/tests/types/test_choice.py
+++ b/tests/types/test_choice.py
@@ -209,3 +209,71 @@ class TestEnumType:
         session.commit()
 
         assert order_nullable.status is None
+
+
+class TestChoiceTypeFalsyCode:
+    """A list-of-tuples ``ChoiceType`` must coerce *every* non-NULL stored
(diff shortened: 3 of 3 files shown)


sqlalchemy_utils/types/choice.py
```diff
@@ -183,7 +183,7 @@ def process_bind_param(self, value, dialect):
         return value
 
     def process_result_value(self, value, dialect):
-        if value:
+        if value is not None:
             return Choice(value, self.choices_dict[value])
         return value
```
kurtmckee: Your patch is very targeted and I appreciate that! However, given that the final line is now guaranteed to return `None`, I'd like to reverse this condition and make the return types explicit (`None` or `Choice`):


```suggestion
        if value is None:
            return None
        return Choice(value, self.choices_dict[value])
```

tests/types/test_choice.py
```diff
@@ -209,3 +209,71 @@ def test_order_nullable_stores_none(self, session, OrderNullable):
...
+    """A list-of-tuples ``ChoiceType`` must coerce *every* non-NULL stored
+    code back into a :class:`Choice`, exactly like the ``Enum`` variant does.
+
+    A falsy-but-valid code (integer ``0`` or the empty string) used to be
+    returned verbatim from the database because ``process_result_value``
+    tested ``if value:`` instead of ``if value is not None:``. The ``Enum``
+    sibling already guards with ``is None`` (see
+    ``TestEnumType.test_setting_value_that_resolves_to_none``), so the two
+    implementations disagreed on the shared contract.
+    """
```
kurtmckee: This provides context that's relevant for this PR, but I don't think the second paragraph is needed in this docstring. I recommend:


```suggestion
    """A list-of-tuples ``ChoiceType`` must coerce *every* non-NULL value
    back into a :class:`Choice`, including false-y values like ``0``
    or the empty string.
    """
```

I see that we spell "falsy" / "false-y" differently, and that's not important to me here. I'm simply recommending that the description of past behavior, historical context, and comparisons to `Enum` and other tests be removed.

tests/types/test_choice.py
```diff
@@ -209,3 +209,71 @@ def test_order_nullable_stores_none(self, session, OrderNullable):
...
+    def test_falsy_code_roundtrips_through_database(self, session, Item):
+        item = Item()
+        item.status = 0
+
+        session.add(item)
+        session.commit()
+
+        item = session.query(Item).first()
+        assert isinstance(item.status, Choice)
+        assert item.status.code == 0
```
kurtmckee, line 272: Because this is testing false-y values, I'd like to see a confirmation that `item.status.code` is an integer (and, specifically, NOT a boolean, which is a subclass of `int`, which precludes using `isinstance()`). I recommend something like:

```python
assert type(item.status.code) is int  # must be an int, not a bool
```

CHANGES.rst
```diff
@@ -3,6 +3,11 @@ Changelog
 
 Here you can see the full list of changes between each SQLAlchemy-Utils release.
 
+Unreleased
+^^^^^^^^^^
+
+- Fix ``ChoiceType`` returning the raw scalar instead of a ``Choice`` for falsy codes such as ``0`` or the empty string (#813)
```
kurtmckee: I think that adding one more sentence will help clarify the behavior here, since `None` is also a false-y value.

```suggestion
- Fix ``ChoiceType`` returning the raw scalar instead of a ``Choice`` for falsy codes such as ``0`` or the empty string. (#813)

  NULL values continue to return ``None``.
```

kurtmckee (changes requested):

binggao1230:
Thanks for the review. Applied all four suggestions in e4921bd: the explicit `if value is None: return None` guard, the trimmed test docstring, the `type(item.status.code) is int` assertion, and the changelog note about NULL. Tests still green.

kurtmckee (approved):

